PR #3, opened 2020-02-14
add huawei cloud kms wrapper

zengchen1024:


zengchen1024:
@jefferai Could you take a look at this pr. Thanks very much!!!

wrappers/huaweicloudkms/huaweicloudkms.go
```diff
@@ -0,0 +1,337 @@
...
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/huaweicloud/golangsdk"
+	huaweisdk "github.com/huaweicloud/golangsdk/openstack"
+	kmsKeys "github.com/huaweicloud/golangsdk/openstack/kms/v1/keys"
+)
+
+// These constants contain the accepted env vars; the Vault one is for backwards compat
+const (
+	EnvHuaweiCloudKMSWrapperKeyID   = "HUAWEICLOUDKMS_WRAPPER_KEY_ID"
+	EnvVaultHuaweiCloudKMSSealKeyID = "VAULT_HUAWEICLOUDKMS_SEAL_KEY_ID"
```
jefferai: You don't need the VAULT one as this never existed directly in Vault.

wrappers/huaweicloudkms/huaweicloudkms.go
```diff
@@ -0,0 +1,337 @@
...
+func (k *Wrapper) SetConfig(config map[string]string) (map[string]string, error) {
+	if config == nil {
+		config = map[string]string{}
+	}
+
+	// Check and set KeyID
+	keyID, err := getConfig(
+		"kms_key_id",
+		os.Getenv(EnvHuaweiCloudKMSWrapperKeyID),
+		os.Getenv(EnvVaultHuaweiCloudKMSSealKeyID),
```
jefferai: Same comment

wrappers/huaweicloudkms/huaweicloudkms.go
```diff
@@ -0,0 +1,337 @@
...
+}
+
+func getConfig(name string, values ...string) (string, error) {
+	for _, v := range values {
+		if "" != v {
+			return v, nil
+		}
+	}
+
+	return "", fmt.Errorf("'%v' not found for HuaweiCloud KMS wrapper configuration", name)
```
jefferai: nitpick: this should be `%s`, not `%v`

wrappers/huaweicloudkms/huaweicloudkms.go
```diff
@@ -0,0 +1,337 @@
...
+const (
+	EnvHuaweiCloudKMSWrapperKeyID   = "HUAWEICLOUDKMS_WRAPPER_KEY_ID"
+	EnvVaultHuaweiCloudKMSSealKeyID = "VAULT_HUAWEICLOUDKMS_SEAL_KEY_ID"
+)
+
+// Wrapper is a Wrapper that uses HuaweiCloud's KMS
+type Wrapper struct {
+	client       kmsClient
+	keyID        string
+	currentKeyID *atomic.Value
```
jefferai: I'm not sure about this currentKeyID mechanism. If this value can be changed on every encrypt operation, then it seems like any time you do a KeyID() call, this value might have been overwritten multiple times. In other words, I'm not sure it's meaningful to display the current actual value as opposed to the original alias.

zengchen1024: Other wrappers also change the `currentKeyID` every time the encrypt operation happens. See [aws](https://github.com/hashicorp/go-kms-wrapping/blob/master/wrappers/awskms/awskms.go#L187-L194). Actually, the key id will not change for huawei cloud kms encryption operation. In other words, the input key id is same as the one the encrypt api returns.

wrappers/huaweicloudkms/huaweicloudkms.go
zengchen1024: Done

wrappers/huaweicloudkms/huaweicloudkms.go
zengchen1024: Done

wrappers/huaweicloudkms/huaweicloudkms.go
zengchen1024: Done

zengchen1024:
@jefferai Could you take a look again. Very appreciate for your reviewing!

wrappers/huaweicloudkms/huaweicloudkms.go
jefferai: In that case, why not just remove this mechanism entirely and return the input key id?

jefferai: Never mind, this is fine (see https://github.com/hashicorp/go-kms-wrapping/pull/8/files)

jefferai:
LGTM, thanks!

wrappers/huaweicloudkms/huaweicloudkms.go
zengchen1024: Thanks!

PR #6, opened 2020-03-06
wrapper/tencentcloudkms: add tencent cloud kms wrapper

likexian:
Hello,

This PR add supports of TencentCloud KMS.

## About TencentCloud
TencentCloud is a secure, reliable and high-performance cloud compute service provided by Tencent.  It is the 2st largest Cloud Provider in China.


likexian:
Hello @jefferai 
Would you please have a review?

wrappers/tencentcloudkms/tencentcloudkms.go
```diff
@@ -0,0 +1,252 @@
...
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common"
+	"github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common/profile"
+	kms "github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/kms/v20190118"
+)
+
+// These constants contain the accepted env vars; the Vault one is for backwards compat
+const (
+	EnvTencentCloudKMSWrapperKeyID   = "TENCENTCLOUDKMS_WRAPPER_KEY_ID"
+	EnvVaultTencentCloudKMSSealKeyID = "VAULT_TENCENTCLOUDKMS_SEAL_KEY_ID"
```
jefferai: This is unnecessary; it never existed in Vault.

wrappers/tencentcloudkms/tencentcloudkms.go
```diff
@@ -0,0 +1,252 @@
...
+// * Instance metadata role (access key and secret key)
+func (k *Wrapper) SetConfig(config map[string]string) (map[string]string, error) {
+	if config == nil {
+		config = map[string]string{}
+	}
+
+	switch {
+	case os.Getenv(EnvTencentCloudKMSWrapperKeyID) != "":
+		k.keyID = os.Getenv(EnvTencentCloudKMSWrapperKeyID)
+	case os.Getenv(EnvVaultTencentCloudKMSSealKeyID) != "":
```
jefferai: Unnecessary, please remove

wrappers/tencentcloudkms/tencentcloudkms.go
```diff
@@ -0,0 +1,252 @@
...
+
+// Wrapper is a wrapper that uses TencentCloud KMS
+type Wrapper struct {
+	accessKey    string
+	secretKey    string
+	sessionToken string
+	region       string
+
+	keyID        string
+	currentKeyID *atomic.Value
```
jefferai: I'm not sure about this currentKeyID mechanism. If this value can be changed on every encrypt operation, then it seems like any time you do a KeyID() call, this value might have been overwritten multiple times. In other words, I'm not sure it's meaningful to display the current actual value as opposed to the original alias.

likexian: Hello @jefferai , thanks for your reviews.

For tencentcloud kms, every encrypt operation will returns a unique `encrypt key id`, which I stores it to `currentKeyID`, this `encrypt key id` is totally different from the `kms key id` that setting in the configure. When do a `KeyID()` call, it just returns the `encrypt key id`, it is a read function.

likexian:
I have make the requested changes, the tests passed.
Please help have a new review.

```
=== RUN   TestAccTencentCloudKMSWrapper_Lifecycle
--- PASS: TestAccTencentCloudKMSWrapper_Lifecycle (0.29s)
=== RUN   TestTencentCloudKMSWrapper
--- PASS: TestTencentCloudKMSWrapper (0.00s)
=== RUN   TestTencentCloudKMSWrapper_Lifecycle
--- PASS: TestTencentCloudKMSWrapper_Lifecycle (0.00s)
PASS
ok      github.com/hashicorp/go-kms-wrapping/wrappers/tencentcloudkms   0.297s
```

wrappers/tencentcloudkms/tencentcloudkms.go
jefferai: That's my point though. The KeyID call will return different values over time, but that is not necessarily the value you're expecting. Let's say you have multiple goroutines using this wrapper. Routine A does an encrypt, then B does an encrypt. Then A reads KeyID but what it gets back isn't the key ID that corresponded to the encrypt call it performed, but instead it gets the key ID corresponding to B's encryption call.

jefferai: Are you sure _every_ encrypt operation returns a _unique_ key ID? As opposed to the key changing over time due to e.g. periodic rotation?

likexian: Yes, every encrypt operation uses a golbal `kms key id` and a unique `encrypt key id`. You may image that every encrypt operation uses a key (`kms key id`) and a salt (`encrypt key id`).

likexian: > That's my point though. The KeyID call will return different values over time, but that is not necessarily the value you're expecting. Let's say you have multiple goroutines using this wrapper. Routine A does an encrypt, then B does an encrypt. Then A reads KeyID but what it gets back isn't the key ID that corresponded to the encrypt call it performed, but instead it gets the key ID corresponding to B's encryption call.

So, what is your suggestions? remove the function `KeyID()` ?

likexian: > Are you sure _every_ encrypt operation returns a _unique_ key ID? As opposed to the key changing over time due to e.g. periodic rotation?

Sorry, I correct what I said: tencentcloud always returns the fixed `kms key id`, what I said below is wrong.

wrappers/tencentcloudkms/tencentcloudkms.go
```diff
@@ -0,0 +1,244 @@
...
+		if keyInfo.Response.KeyMetadata == nil || keyInfo.Response.KeyMetadata.KeyId == nil {
+			return nil, fmt.Errorf("no key information return")
+		}
+
+		k.currentKeyID.Store(*keyInfo.Response.KeyMetadata.KeyId)
+		k.client = client
+	}
+
+	wrappingInfo := make(map[string]string)
+	wrappingInfo["retgion"] = k.region
```
jefferai: typo

wrapper.go
```diff
@@ -8,17 +8,18 @@ import (
...
-	Test           = "test-auto"
+	AEAD            = "aead"
+	Shamir          = "shamir"
+	PKCS11          = "pkcs11"
+	AliCloudKMS     = "alicloudkms"
+	AWSKMS          = "awskms"
+	GCPCKMS         = "gcpckms"
+	AzureKeyVault   = "azurekeyvault"
+	OCIKMS          = "ocikms"
+  HuaweiCloudKMS = "huaweicloudkms"
```
jefferai: Formatting changed for some reason

wrappers/tencentcloudkms/tencentcloudkms.go
likexian: Sorry!

wrapper.go
likexian: Formatting changed due to github conflict merge, fixed.
The tests of HuaweiCloudKMS never passed, I also fixed it.

likexian:
> I'm going to let it go this time but in the future, please do not lint/adjust things that are not in scope of the PR.
> 
> LGTM!

Sorry, got it.
I fixed it because the global `go test` failed.

PR #9, opened 2020-04-07
Add structwrapping package, which can wrap structs automatically via tags

jefferai:


structwrapping/README.md
```diff
@@ -0,0 +1,53 @@
...
+Caveats:
+* The input must be a pointer to a struct
+* Tags must be balanced (there must be a matching ct tag entry with the same
+  identifier as a pt tag entry, and vice versa)
+* This does not currently recurse
+* Currently there is no checking for overwriting of values, although this may
+  change
+
+## Usage
+
```
jimlambrt, line 19:
```suggestion
Add struct tags that specify fields for wrapping.  Wrapping tags must contain:
* a field type of pt (plaintext) or ct (ciphertext) 
* a unique name that ties together a pair of pt and ct fields 

```

structwrapping/structwrapping.go
```diff
@@ -0,0 +1,154 @@
...
+		if v[1] == nil {
+			return nil, fmt.Errorf("no ct wrapping tag found for identifier %q", k)
+		}
+	}
+
+	return edMap, nil
+}
+
+func WrapStruct(ctx context.Context, wrapper wrapping.Wrapper, in interface{}) error {
+	if wrapper == nil {
```
jimlambrt, line 97: perhaps check ctx for nil?  if you ever decide to use the ctx in the future, you could have a lot broken clients that have been passing nil

structwrapping/structwrapping_test.go
```diff
@@ -0,0 +1,165 @@
...
+	"testing"
+
+	wrapping "github.com/hashicorp/go-kms-wrapping"
+	"github.com/hashicorp/go-kms-wrapping/wrappers/aead"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestStructWrapping(t *testing.T) {
+	wrapper := aead.NewWrapper(nil)
+	wrapper.SetConfig(map[string]string{"aead_type": "aes-gcm", "key": "QmjueU/LMsZAO+rvSUMcpkBziCD5ON7BgxVqcZ6+TCI="})
```
jimlambrt, line 13: perhaps check for err here... it might save a future dev some time tracking down why the tests are failing.

structwrapping/structwrapping_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	t.Run("shared encryption/decryption tests", func(t *testing.T) {
+		t.Run("bad basic values", func(t *testing.T) {
```
jimlambrt, line 16: I always forget to use subtests like this.  ty for the reminder.

structwrapping/structwrapping.go
jefferai: To be honest, I'm not sure how libraries writ large, by which I mean Go stdlib and others, handle nil ctxes. Considering it's a pure passthrough though I don't really feel like this is the place to take a stand.

structwrapping/structwrapping_test.go
jefferai: Done

structwrapping/structwrapping_test.go
jefferai: :-)

PR #25, opened 2020-09-21
Adds a new interface that includes methods for lifecycle management of keys in a KMS

austingebauer:
## Overview

This PR adds a new interface, `LifecycleWrapper`, which is a `Wrapper` that includes interface methods for lifecycle management of keys in a KMS.

A `KeyNotRequired` field has also been added to `WrapperOptions` in order to provide consumers the option to not provide an existing key when configuring a `Wrapper`. This option is useful for those who want to use `LifecycleWrapper` methods without the need of providing an existing key for cryptographic operations. `KeyNotRequired` has a default of false in order to ensure that key validation behavior continues for existing consumers of the library.

## Testing

I tested this change manually by taking the following steps:

1. Bringing this branch into Vault via `go get`
2. Successfully compiling Vault
3. Configuring an `azurekeyvault` seal
4. Successfully auto-unsealing Vault with the `azurekeyvault` seal

Output from manual testing of this change is below:
```sh
$ git branch --show-current
master

$ go get github.com/hashicorp/go-kms-wrapping@key-lifecycle-wrapper

$ make dev
==> Checking that build is using go version >= 1.14.7...
==> Using go version 1.14.7...
==> Removing old directory...
==> Building...
Number of parallel builds: 15

-->    darwin/amd64: github.com/hashicorp/vault

==> Results:
total 258264
-rwxr-xr-x  1 austingebauer  staff   126M Sep 21 15:42 vault

$ cat config/ha1.hcl 
storage "consul" {}
listener "tcp" {
  address = "127.0.0.1:8200"
  tls_disable = true
}

seal "azurekeyvault" {
  tenant_id      = "<redacted>"
  client_id      = "<redacted>"
  client_secret  = "<redacted>"
  vault_name     = "keyvault-keys-test"
  key_name       = "unseal"
}

$ vault server -log-level=trace -config config/ha1.hcl > /tmp/s1.log 2>&1 &
[1] 11740

$ vault operator init -recovery-shares 1 -recovery-threshold 1 
Recovery Key 1: 1MCS4S6SS4IAdxFocvqbcVgzYaXIWBoHn1jk+k2LlYI=

Initial Root Token: s.H1aNRwOTC0Qwb5TXFyJYLcpS

Success! Vault is initialized

Recovery key initialized with 1 key shares and a key threshold of 1. Please
securely distribute the key shares printed above.

$ vault status
Key                      Value
---                      -----
Recovery Seal Type       shamir
Initialized              true
Sealed                   false
Total Recovery Shares    1
Threshold                1
Version                  1.5.0
Storage Type             consul
Cluster Name             vault-cluster-23b8e140
Cluster ID               336573dd-492d-db31-04f8-9843cb02e028
HA Enabled               true
HA Cluster               https://127.0.0.1:8201
HA Mode                  active
```

### DIFF
--- a/wrapper.go
+++ b/wrapper.go
@@ -42,9 +42,12 @@ const (
 type Purpose string
 
 const (
-	EncryptDecrypt Purpose = "encrypt_decrypt"
-	SignVerify     Purpose = "sign_verify"
-	WrapUnwrap     Purpose = "wrap_unwrap"
+	Encrypt Purpose = "encrypt"
+	Decrypt Purpose = "decrypt"
+	Sign    Purpose = "sign"
+	Verify  Purpose = "verify"
+	Wrap    Purpose = "wrap"
+	Unwrap  Purpose = "unwrap"
 )
 
 // ProtectionLevel defines where cryptographic operations are performed with a key.
@@ -58,7 +61,7 @@ const (
 // KMSKey represents a cryptographic key that can be imported into a KMS.
 type KMSKey struct {
 	Type            KeyType
-	Purpose         Purpose
+	Purposes        []Purpose
 	ProtectionLevel ProtectionLevel
 	Material        KeyMaterial
 }
--- a/wrapper.go
+++ b/wrapper.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/vault/sdk/helper/keysutil"
 )
 
 // These values define known types of Wrappers
@@ -28,6 +29,23 @@ const (
 	HSMAutoDeprecated = "hsm-auto"
 )
 
+// Purpose defines the cryptographic capabilities of a key.
+type Purpose string
+
+const (
+	EncryptDecrypt Purpose = "encrypt_decrypt"
+	SignVerify     Purpose = "sign_verify"
+	WrapUnwrap     Purpose = "wrap_unwrap"
+)
+
+// ProtectionLevel defines where cryptographic operations are performed with a key.
+type ProtectionLevel string
+
+const (
+	Software ProtectionLevel = "software"
+	HSM      ProtectionLevel = "hsm"
+)
+
 // Wrapper is the embedded implementation of autoSeal that contains logic
 // specific to encrypting and decrypting data, or in this case keys.
 type Wrapper interface {
@@ -50,7 +68,36 @@ type Wrapper interface {
 	Decrypt(context.Context, *EncryptedBlobInfo, []byte) ([]byte, error)
 }
 
+// LifecycleWrapper is a Wrapper that implements lifecycle management for keys in a KMS.
+type LifecycleWrapper interface {
+	Wrapper
+
+	// ImportKey creates a named key by importing key material in the given KeyEntry.
+	// The key will have the given KeyType, Purpose, and ProtectionLevel if supported by the implementation.
+	// Returns the ID of a new key version and an error.
+	ImportKey(ctx context.Context, name string, kt keysutil.KeyType, ke keysutil.KeyEntry, pr Purpose, pl ProtectionLevel) (string, error)
+
+	// RotateKey rotates the named key by creating a new key version with key material in the given KeyEntry.
+	// The key version will have the given KeyType, Purpose, and ProtectionLevel if supported by the implementation.
+	// Returns the ID of a new key version and an error.
+	RotateKey(ctx context.Context, name string, kt keysutil.KeyType, ke keysutil.KeyEntry, pr Purpose, pl ProtectionLevel) (string, error)
+
+	// DeleteKey deletes the named key.
+	// Returns a bool representing if the key exists and an error.
+	DeleteKey(ctx context.Context, name string) (bool, error)
+
+	// EnableKeyVersion enables the version of the named key.
+	EnableKeyVersion(ctx context.Context, name, version string) error
+
+	// DisableKeyVersion disables the version of the named key.
+	DisableKeyVersion(ctx context.Context, name, version string) error
+}
+
 // WrapperOptions contains options used when creating a Wrapper
 type WrapperOptions struct {
 	Logger hclog.Logger
+
+	// KeyNotRequired indicates if an existing key must be
+	// supplied in the configuration for a Wrapper.
+	KeyNotRequired bool
 }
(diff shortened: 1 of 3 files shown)


wrapper.go
```diff
@@ -50,7 +68,36 @@ type Wrapper interface {
...
+
+	// ImportKey creates a named key by importing key material in the given KeyEntry.
+	// The key will have the given KeyType, Purpose, and ProtectionLevel if supported by the implementation.
+	// Returns the ID of a new key version and an error.
+	ImportKey(ctx context.Context, name string, kt keysutil.KeyType, ke keysutil.KeyEntry, pr Purpose, pl ProtectionLevel) (string, error)
+
+	// RotateKey rotates the named key by creating a new key version with key material in the given KeyEntry.
+	// The key version will have the given KeyType, Purpose, and ProtectionLevel if supported by the implementation.
+	// Returns the ID of a new key version and an error.
+	RotateKey(ctx context.Context, name string, kt keysutil.KeyType, ke keysutil.KeyEntry, pr Purpose, pl ProtectionLevel) (string, error)
```
tvoran, line 83: The args list looks a little long for ImportKey and RotateKey, I wonder if passing a struct here would make sense instead. Just a thought.

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -253,6 +262,26 @@ func (v *Wrapper) Decrypt(ctx context.Context, in *wrapping.EncryptedBlobInfo, a
...
+}
+
+func (v *Wrapper) EnableKeyVersion(_ context.Context, _, _ string) error {
+	return nil
+}
+
+func (v *Wrapper) DisableKeyVersion(_ context.Context, _, _ string) error {
+	return nil
+}
+
```
tvoran, line 284: Are these intentionally not implemented?

wrapper.go
austingebauer: I agree. I thought about passing a struct for these as well. Happy to change this to use a struct if others agree!

wrappers/azurekeyvault/azurekeyvault.go
austingebauer: Yep, it's intentional. I wanted to keep this PR focused on the interface. The implementations will come next.

wrapper.go
calvn: Depending on how the implementation consumes `KeyEntry`, this might also be an opportunity to remove the dependency on the `keysutil` package. The struct could contain all the information needed for the operation, such as the key type, key bytes, purpose, protection level, etc.

wrapper.go
```diff
@@ -50,7 +68,36 @@ type Wrapper interface {
...
 }
 
+// LifecycleWrapper is a Wrapper that implements lifecycle management for keys in a KMS.
+type LifecycleWrapper interface {
+	Wrapper
+
+	// ImportKey creates a named key by importing key material in the given KeyEntry.
+	// The key will have the given KeyType, Purpose, and ProtectionLevel if supported by the implementation.
+	// Returns the ID of a new key version and an error.
+	ImportKey(ctx context.Context, name string, kt keysutil.KeyType, ke keysutil.KeyEntry, pr Purpose, pl ProtectionLevel) (string, error)
```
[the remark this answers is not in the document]
calvn: I thought this was the case, but I'm mistaken.

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -118,6 +123,8 @@ func (v *Wrapper) SetConfig(config map[string]string) (map[string]string, error)
 	}
 
 	switch {
+	case v.keyNotRequired:
```
calvn, line 126: Can there be a case where a key is not required but provided anyways such that it's supposed to be used? If so, we might want to move this towards the bottom of the switch, right above the default case.

austingebauer: Yep, I could see there being a case for that. I've moved it as you've suggested in a34f78c.

wrapper.go
austingebauer: I like the idea of removing the `keysutil` package dependency. I've changed `ImportKey` and `RotateKey` to take in a newly defined struct, `KMSKey`, which contains all of the information necessary for those operations. Please have a look at let me know your thoughts.

tvoran (approved):

wrapper.go
```diff
@@ -28,6 +29,48 @@ const (
 	HSMAutoDeprecated = "hsm-auto"
 )
 
+// KeyType defines types of cryptographic keys.
+type KeyType string
+
+const (
+	RSA2048 KeyType = "rsa_2048"
```
calvn, line 36: Any reason why we're going with `rsa_2048` for the string value representation (as opposed to `rsa-2048` or `rsa2048`)? Keep in mind that these are shared across provider implementations, so what's likely to work for one provider may not for work for another if these are intended to be passed as string values to the provider's SDK APIs. Same for purpose and protection level.

wrapper.go
```diff
@@ -28,6 +29,48 @@ const (
...
+type KMSKey struct {
+	Type            KeyType
+	Purposes        []Purpose
+	ProtectionLevel ProtectionLevel
+	Material        KeyMaterial
+}
+
+// KeyMaterial contains key material for various key types.
+type KeyMaterial struct {
+	RSAKey *rsa.PrivateKey
```
calvn, line 71: Is this the only bit of information needed for key lifecycle operations that require key material?

calvn (approved):
Left some questions, but the interface looks good!

wrapper.go
austingebauer: There's no particular reason the string value is `rsa_2048` for `KeyType`. These constants are meant to be generic and work across different providers. The intent is for the implementations of `LifecycleWrapper` methods to map the 
 `KeyType` to a value that's expected by the specific provider's SDK APIs. Taking Azure Key Vault as an example, the `RSA2048` `KeyType` would map to one of [these values](https://github.com/Azure/azure-sdk-for-go/blob/master/services/keyvault/v7.0/keyvault/models.go#L173). The same is true for purpose and protection level.

wrapper.go
austingebauer: That's correct. Importing and rotating only require an `*rsa.PrivateKey` now. When more key types become supported, they'll be added to this struct.

PR #35, opened 2021-07-17
gcpkms: add client getter and option to not provide crypto key in config

austingebauer:
This PR adds usage of the `WrapperOptions.KeyNotRequired` option to the GCP KMS wrapper. Current clients of the wrapper will not experience a behavior change, as the default value requires the crypto key to be provided. The change in this PR is similar to what was done for the [Azure Key Vault](https://github.com/hashicorp/go-kms-wrapping/blob/master/wrappers/azurekeyvault/azurekeyvault.go#L61) and [AWS KMS](https://github.com/hashicorp/go-kms-wrapping/blob/master/wrappers/awskms/awskms.go#L63) wrappers.

This PR also adds a getter for the client used by the GCP KMS wrapper.

### DIFF
--- a/wrappers/gcpckms/gcpckms.go
+++ b/wrappers/gcpckms/gcpckms.go
@@ -47,7 +47,8 @@ type Wrapper struct {
 
 	userAgent string
 
-	currentKeyID *atomic.Value
+	currentKeyID   *atomic.Value
+	keyNotRequired bool
 
 	client *cloudkms.KeyManagementClient
 }
@@ -59,7 +60,8 @@ func NewWrapper(opts *wrapping.WrapperOptions) *Wrapper {
 		opts = new(wrapping.WrapperOptions)
 	}
 	s := &Wrapper{
-		currentKeyID: new(atomic.Value),
+		currentKeyID:   new(atomic.Value),
+		keyNotRequired: opts.KeyNotRequired,
 	}
 	s.currentKeyID.Store("")
 	return s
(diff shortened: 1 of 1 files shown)


calvn (approved):

jefferai:
Can this wait until v2 is ready so it doesn't need to be ported? Or is there urgency here?

austingebauer:
@jefferai - Do you have an idea of timing for v2? It'd be helpful to get this merged so that work on the ent side can proceed. I don't mind porting this to v2.

jefferai:
My guess is a month-ish. A lot of the groundwork is laid, the main thing left is working on the plugin model which will hopefully not be too bad. But then the individual wrappers have to be ported. And that month estimate is not including however long it takes for Vault to switch over.

So I guess let's do it here and I'll port it over after. It may not be bad, I'm assuming a straight merge wouldn't make things happy, but the wrapper code hasn't really changed (other than `aead`) so it might not be bad!

jefferai:
One note on the code, which isn't really related to this review but...

Assuming it's allowed for SetConfig to be called more than once, s.client should probably be an atomic.Value, similar to the current key ID.

Generally speaking we sort of assume that SetConfig won't be called multiple times but it's a public function and I don't think it's documented that way. A wrapper that specifically doesn't want to allow SetConfig to be called multiple times might want to error out if it is.

jefferai (approved):

PR #41, opened 2021-08-24
Base module v2

jefferai:
This makes the following changes to the base module of go-kms-wrapping:

* Renames various things for better proto compatibility (IV -> Iv, HMAC
  -> Hmac, XOR-> Xor, etc.) Naming is debatable, and this doesn't match
  "official" Go conventions, but...not everyone thinks those conventions
  are best.  After all, GCPCKMS screams at you less when it's GcpCkms.
  (Proto generates this case by default; Vault forces things back to
  "normal" Go via Makefile commands.)
* Defines some new consts for AEAD and hash function types
* Introduces an Option pattern for optional parameters. This allows for
  updating capabilities of functions without breaking API. It's built in
  such a way that we can standardize in wrappers on `wrapping.Option`
  but still be able to know which package should parse it.
* Formalizes the SetConfig pattern as a part of the main wrapper. This
  allows setting opaque configuration without having to pull in the
  wrapper package itself (useful for plugins)
* Deprecates the Vault-specific "Wrapped" boolean in favor of a
  "plaintext" member. That way both cipher- and plain-text can be
  carried in a message and used as appropriate instead of relying on a
  bool to indicate which the "ciphertext" bytes refer to.
* Deprecates the Vault-specific "ValuePath" string in favor of an opaque
  ClientData structure, which can include the path as part of its data.
* Renames EncryptedBlobInfo to just BlobInfo as it...may not be
  encrypted.
* Moves `multiwrapper` into the main module. Since wrappers are going to
  start being their own modules, it didn't make sense to keep it under
  `wrappers/`. Note that the tests for multiwrapper and structwrapping
  are moved into the `aead` module which will come in a future PR. I may
  move them back at some point if I switch those to use the test wrapper
  instead of AEAD but the benefit of AEAD is that some values like AAD
  can be tested there.
* Splits the Wrapper interface into three that can be satisfied
  independently: Wrapper, InitFinalizer, and HmacComputer. This way if a
  wrapper doesn't actually _do_ anything for initialization/finalization
  it doesn't need to have those invoked "just in case" which can lead to
  some timing issues in downstream apps.
* Adds implementations in testing.go for the various interfaces, as well
  as some combinations.
* Updates all interface members to take in contexts and return errors.
  When directly invoking a wrapper some methods may not look at the
  context and always return nil, but since we're supporting a plugin
  interface too it's better to allow errors from the plugin layer to
  surface rather than swallow them.


jefferai:
I recommend reviewing `wrapping.go` first followed by the `proto` file, then `options.go`. After that things mostly fall into place.

sgmiller (approved):
Looks good to me.  A couple of these will help the PKCS#11 backed PKI secrets engine effort as well, where we introduced on options pattern as well.  Will land on this one later.

README.md
```diff
@@ -51,13 +51,13 @@ as they may have been used for past encryption operations.
   * Supports Additional Authenticated Data (AAD) for all KMSes except Vault Transit.
 
 A
-[`multiwrapper`](https://github.com/hashicorp/go-kms-wrapping/tree/master/wrappers/multiwrapper)
+[`multiwrapper`](https://github.com/hashicorp/go-kms-wrapping/tree/v2/wrappers/multiwrapper)
```
calvn, line 54:
```suggestion
[`multiwrapper`](https://github.com/hashicorp/go-kms-wrapping/tree/v2/multiwrapper)
```

Should it be this instead since we're moving it up?

calvn (approved):
Changes look good, big fan of the change to use optional parameters! Left a small comment around the go version in `go.mod`.

README.md
jefferai: Thanks!

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/hashicorp/go-kms-wrapping
+module github.com/hashicorp/go-kms-wrapping/v2
 
 go 1.13
```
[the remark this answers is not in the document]
jefferai: Done, thanks!

jimlambrt (approved):
I've reviewed this along the way.  ty!

github.com.hashicorp.go.kms.wrapping.v2.types.proto
```diff
@@ -0,0 +1,95 @@
...
+	bytes iv = 2;
+
+	// HMAC is the bytes of the HMAC, if any
+	bytes hmac = 3;
+
+	// Wrapped can be used by the client to indicate whether Ciphertext actually
+	// contains wrapped data or not. This can be useful if you want to reuse the
+	// same struct to pass data along before and after wrapping. Deprecated in
+	// favor of plaintext.
+	bool wrapped = 4 [deprecated = true];
```
briankassouf, line 44: This will be a go.mod v2 module right? If so, do we want to remove instead of deprecate?

github.com.hashicorp.go.kms.wrapping.v2.types.proto
```diff
@@ -0,0 +1,95 @@
...
+// Options holds options common to all wrappers
+message Options {
+	// The key ID being specified
+	string with_key_id = 10;
+
+	// The AAD bytes, if any
+	bytes with_aad = 20;
+
+	// Wrapper-specific options to pass along
+	map<string, string> with_wrapper_options = 30;
```
briankassouf, line 94: Are we expecting to slot in a lot of options between 10, 20, & 30? What order are you hoping to encode here? 

Apparently field numbers > 16 take 2 bytes to encode, so if these will be frequently encoded and stored we may want to avoid higher numbers if we can. I'm not sure that's the case for this options type though, right? So may not matter.

multiwrapper/multiwrapper.go
```diff
@@ -0,0 +1,193 @@
...
+var _ wrapping.Wrapper = (*MultiWrapper)(nil)
+
+var ErrKeyNotFound = errors.New("given key ID not found")
+
+// MultiWrapper allows multiple wrappers to be used for decryption based on key
+// ID. This allows for rotation of data by allowing data to be decrypted across
+// multiple (possibly derived) wrappers and encrypted with the default.
+// Functions on this type will likely panic if the wrapper is not created via
+// NewMultiWrapper.
+type MultiWrapper struct {
```
briankassouf, line 23: Very pedantic, but i was a bit confused by this name. I expected this to encrypt a plaintext N times with N different wrappers. Perhaps `MultiUnwrapper`?

briankassouf (approved):
Looks great, all my comments are very optional

github.com.hashicorp.go.kms.wrapping.v2.types.proto
jefferai: This is for Vault. Keeping it in but deprecated means that it's a part of the resulting struct, so you can unwrap v1-encoded structs with the v2 struct (I have a test for this in Boundary that encrypts in v1 and decrypts in v2 and vice versa, although it needs to be uncommented to run so as not to pull in both versions as deps). Eventually if Vault is no longer using this bool (instead using plaintext/ciphertext fields) then we can remove and reserve the number.

github.com.hashicorp.go.kms.wrapping.v2.types.proto
jefferai: This was just for organizational neatness. It means if we want to slot in new fields we can put them in a place that logically makes sense when looking at the proto and have the numbers strictly incrementing.

If you like I can avoid doing this. I don't think a 1-byte penalty is all that bad; even if you are storing a billion entries and you are talking about a gigabyte that will still be a tiny fraction of the overall amount...but not nothing.

Let me know your preference.

multiwrapper/multiwrapper.go
jefferai: I'll rename it.

jefferai: I think maybe what I'll do is change the module to `multi` and change this to `PooledWrapper`. I think your idea of encrypting multiple times with multiple wrappers is interesting and from that perspective may want to reserve MultiWrapper/MultiUnwrapper names for that purpose. Sound good?

github.com.hashicorp.go.kms.wrapping.v2.types.proto
briankassouf: Ah ok, this makes sense, thanks!

github.com.hashicorp.go.kms.wrapping.v2.types.proto
briankassouf: No strong preference. I don't think these will realistically be stored ever, so let's not worry about it

multiwrapper/multiwrapper.go
briankassouf: That name works for me! Thanks

PR #42, opened 2021-09-20
Update to oci-sdk-go 13.1.0 for Qualsys issue

sgmiller:
The Oracle OCI-go library was updated quite a while ago to add support for their London datacenter.  This branch was created to minimally upgrade the library to address a customer issue.  The library is now at 24.x, so we're pretty far behind.  However this addresses the issue for customers of Oracle's London DC.


jefferai (changes requested):
Please `go mod tidy` when you update deps.

jefferai:
Any reason you're going from "super old" to "very old"? The current SDK version is `24.3.0`. I don't know the changes along the way but is there a reason not to just bump all the way up?

sgmiller:
No reason other than to address only this issue and not need to worry about any changes in the library that would break existing functionality.

jefferai (approved):

PR #47, opened 2022-01-28
Always set currentKeyID in SetConfig

rculpepper:
This adds a line to make sure that `currentKeyID` is always set when `SetConfig` is called, so that calls to the `KeyID()` function will return the correct key ID when `keyNotRequired` is `true`.


sgmiller (approved):

austingebauer (approved):

jefferai (approved):

kitography (commented):
Is there a way to write a test case for this?

PR #49, opened 2022-02-16
Add the ability for a Azure/AWS wrapper to ignore environment variables

stevendpclark:
- A minimal change due to the v2 version actively being worked on to allow
   Vault to specify all the configuration parameters to either an Azure or
   AWS wrapper and ignore any environment variables that would be set/override 
   the provided values within the config map.
 - This is to allow the new Managed Key feature within Vault to specify the entire configuration 
   to a wrapper and not have an existing seal wrap environment var override configuration of the 
  managed key for example.
 - Relates to VAULT-5067

### DIFF
--- a/wrappers/awskms/awskms.go
+++ b/wrappers/awskms/awskms.go
@@ -76,19 +76,31 @@ func NewWrapper(opts *wrapping.WrapperOptions) *Wrapper {
 //
 // Order of precedence AWS values:
 // * Environment variable
-// * Value from Vault configuration file
+// * Passed in config map
 // * Instance metadata role (access key and secret key)
 // * Default values
 func (k *Wrapper) SetConfig(config map[string]string) (map[string]string, error) {
+	return k.SetConfigWithEnv(config, true)
+}
+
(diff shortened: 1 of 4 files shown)


divyapola5 (approved):
Looks good to me.
Reminder to link JIRA story :)

cipherboy (approved):

divyapola5 (approved):
👍 for UTs

stevendpclark:
> I'm a bit confused by this TBH. Why are we adding this change if it's always forced on?
> 
> I'd expect that we would instead make it a value in the config map that can be user-set.

I'm assuming that the Managed Key code within Vault would start using the SetConfigWithEnv function passing a value of false to the allowEnv flag. The SetConfig function is there to maintain backwards compatibility. I thought it was a bit clearer and easier way to audit what code is using this, then passing a new value within the wrapper's config map which may/or may not enforce it.

I'm not against what you are suggesting and keeping a singular SetConfig with a config key value of say "allow_env_vars".

jefferai (approved):

cipherboy (approved):

PR #51, opened 2022-02-16
Add plugin module, which allows running wrappers via go-plugin

jefferai:
Note: structwrapping test moved to aead package in a previous PR, hence the deletion.


plugin/github.com.hashicorp.go.kms.wrapping.plugin.v2.proto
```diff
@@ -0,0 +1,76 @@
...
+message TypeRequest {}
+
+message TypeResponse {
+	string type = 10;
+}
+
+message KeyIdRequest {}
+
+message KeyIdResponse {
+	string key_id = 10 [json_name="key_id"];
```
calvn, line 18: Why is the `json_name` option needed for `key_id`?

plugin/github.com.hashicorp.go.kms.wrapping.plugin.v2.proto
```diff
@@ -0,0 +1,76 @@
...
+option go_package = "github.com/hashicorp/go-kms-wrapping/v2/plugin;plugin";
+
+package github.com.hashicorp.go.kms.wrapping.plugin.v2;
+
+import "github.com.hashicorp.go.kms.wrapping.v2.types.proto";
+
+message TypeRequest {}
+
+message TypeResponse {
+	string type = 10;
```
calvn, line 12: How come we're starting the index at 10 (and sometimes 20 for other messages)?

plugin/options.go
```diff
@@ -0,0 +1,52 @@
...
+	return &options{}
+}
+
+// WithLogger allows passing a logger to the plugin library for debugging
+func WithLogger(with hclog.Logger) Option {
+	return func() interface{} {
+		return OptionFunc(func(o *options) {
+			o.withLogger = with
+		})
+	}
```
calvn, line 51: Nit:
```suggestion
func WithLogger(logger hclog.Logger) Option {
	return func() interface{} {
		return OptionFunc(func(o *options) {
			o.withLogger = logger
		})
	}
```

plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
+package plugin
+
+import (
+	context "context"
```
calvn, line 4: Does the import need to be named explicitly here?

plugin/plugin_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	pluginPath := os.Getenv("PLUGIN_PATH")
+	if pluginPath == "" {
+		t.Skipf("skipping plugin test as no PLUGIN_PATH specified")
+	}
+
+	wrapper, cleanup := TestPlugin(t, pluginPath)
+
+	require.NotNil(cleanup)
+	defer cleanup()
```
calvn, line 29: Looks like you do cleanup via `t.Cleanup` in other tests, should we also do it that way here?

plugin/plugin_test.go
```diff
@@ -0,0 +1,127 @@
...
+	ctx := context.Background()
+
+	pluginPath := os.Getenv("PLUGIN_PATH")
+	if pluginPath == "" {
+		t.Skipf("skipping plugin test as no PLUGIN_PATH specified")
+	}
+
+	wrapper, cleanup := TestPlugin(t, pluginPath)
+
+	require.NotNil(cleanup)
```
calvn, line 28: Instead of nil-checking here and elsewhere whenever `TestPlugin` is called, would it be slightly better for TestPlugin to just return a cleanup func that no-ops if there's nothing to clean up?

calvn (commented):
Mostly small comments, and a few questions around the proto file.

plugin/plugin_test.go
```diff
@@ -0,0 +1,127 @@
...
+func TestAeadPluginWrapper(t *testing.T) {
+	require := require.New(t)
+	ctx := context.Background()
+
+	pluginPath := os.Getenv("PLUGIN_PATH")
+	if pluginPath == "" {
+		t.Skipf("skipping plugin test as no PLUGIN_PATH specified")
+	}
+
+	wrapper, cleanup := TestPlugin(t, pluginPath)
```
calvn, line 26: An alternate way to "execute" a plugin without having to compile one before testing would be to have the plugin's main func as a Test func, and make use of `-test.run=SomeWrapperMainFunc` to have go-plugin dispense a client whose command is an execution of the test func itself.

Vault does this via [TestAddTestPlugin](https://github.com/hashicorp/vault/blob/3ff381abe5aa894210146f01bfde362177fa00dd/vault/testing.go#L496). Sample usage can be found [here](https://github.com/hashicorp/vault/blob/3ff381abe5aa894210146f01bfde362177fa00dd/builtin/logical/database/backend_test.go#L49-L67).

plugin/github.com.hashicorp.go.kms.wrapping.plugin.v2.proto
jefferai: See https://developers.google.com/protocol-buffers/docs/proto3 -- some marshalers creating JSON from protobuf will use lowerCamelCase by default for multi-word values, so in these cases I tend to explicitly specify the desired JSON name that aligns with our normal practices.

plugin/github.com.hashicorp.go.kms.wrapping.plugin.v2.proto
jefferai: I tend to leave space so that we can insert more parameters in between later. I'm actually not sure why some start with 20, I've updated all to 10.

If you feel strongly I can start them all at 1.

plugin/options.go
jefferai: This is my general pattern for options (although I've not been 100% consistent in the package -- I will update the base module to fix that later). Since it says "WithX", the parameter is the thing that it's with.

It's just a slightly more standardized way to do these. I don't feel strongly about it, but I also don't see a reason to change it.

plugin/plugin.go
jefferai: No idea why it did that. Some auto-insertion thing. Will remoev.

plugin/plugin_test.go
jefferai: Sure.

plugin/plugin_test.go
jefferai: Sure, good idea.

jefferai: It actually already does that, basically! It can't return with anything to cleanup without it already failing the test. But I'll make it more explicit.

plugin/plugin_test.go
jefferai: I prefer this method, because it aligns much more closely with what's actually going to happen.

plugin/github.com.hashicorp.go.kms.wrapping.plugin.v2.proto
jefferai: Ah I remember -- I just kept `options` as field 20 across all messages for consistency. Just a style, but can use something else if you prefer.

plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+		HandshakeConfig: HandshakeConfig,
+		VersionedPlugins: map[int]gp.PluginSet{
+			1: {"wrapping": &wrapper{}},
+		},
+		Cmd: exec.Command(pluginPath),
+		AllowedProtocols: []gp.Protocol{
+			gp.ProtocolGRPC,
+		},
+		Logger:   opts.withLogger,
+		AutoMTLS: true,
```
briankassouf, line 75: Should we have an option for providing the SHA via `gp.SecureConfig`?

plugin/plugin.go
```diff
@@ -0,0 +1,86 @@
...
+// HandshakeConfig is a shared config that can be used regardless of wrapper, to
+// avoid having to know type-specific things about each plugin
+var HandshakeConfig = gp.HandshakeConfig{
+	MagicCookieKey:   "HASHICORP_GKW_PLUGIN",
+	MagicCookieValue: "wrapper",
+}
+
+// wrapper embeds Plugin and is used as the top-level
+type wrapper struct {
+	gp.Plugin
```
briankassouf, line 23: I don't see this being set anywhere, am i missing it or is it coming in a follow on PR?

briankassouf: Oh, or should this be

```suggestion
	// Embeding this will disable the netRPC protocol
	plugin.NetRPCUnsupportedPlugin
```

briankassouf (approved):

plugin/plugin.go
jefferai: Good thought! I will add that.

jefferai: Added!

jimlambrt (approved):
Looks great.

plugin/github.com.hashicorp.go.kms.wrapping.plugin.v2.proto
```diff
@@ -0,0 +1,76 @@
+syntax = "proto3";
+
+option go_package = "github.com/hashicorp/go-kms-wrapping/v2/plugin;plugin";
+
+package github.com.hashicorp.go.kms.wrapping.plugin.v2;
+
+import "github.com.hashicorp.go.kms.wrapping.v2.types.proto";
+
+message TypeRequest {}
```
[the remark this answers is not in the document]
jefferai: Given that it's exactly mapping to the documented types and interfaces from the main module, I'll add a package level doc comment linking to that (post tagging). It just avoids a lot of "Encrypt satisfies the Encrypt method. Decrypt satisfies the Decrypt method." style bits.

PR #58, opened 2022-03-01
Port huaweicloudkms to v2

jefferai:


jefferai:
@likexian @zengchen1024 Reviews from you would be great

PR #60, opened 2022-03-01
Port tencentcloudkms to v2

jefferai:


jefferai:
@likexian Review from you would be great

likexian:
> @likexian Review from you would be great

lgtm

PR #64, opened 2022-03-15
feature: add NewDerivedReader(...), HmacSha256(...) and HmacSha256Prk(...)

jimlambrt:
replaces #52

### DIFF
--- a/crypto/derived_reader.go
+++ b/crypto/derived_reader.go
@@ -1,0 +1,46 @@
+package crypto
+
+import (
+	"crypto/sha256"
+	"fmt"
+	"io"
+
+	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
+	"golang.org/x/crypto/hkdf"
+)
+
+// DerivedReader returns a reader from which keys can be read, using the
+// given wrapper, reader length limit, salt and context info. Salt and info can
+// be nil.
+//
+// Example:
+//	reader, _ := crypto.NewDerivedReader(wrapper, userId, jobId)
+// 	key := ed25519.GenerateKey(reader)
+func NewDerivedReader(wrapper wrapping.Wrapper, lenLimit int64, salt, info []byte) (*io.LimitedReader, error) {
+	const (
+		op     = "reader.NewDerivedReader"
+		minLen = 20
+	)
+	if wrapper == nil {
+		return nil, fmt.Errorf("%s: missing wrapper: %w", op, wrapping.ErrInvalidParameter)
+	}
+	if lenLimit < minLen {
+		return nil, fmt.Errorf("%s: lenLimit must be >= %d: %w", op, minLen, wrapping.ErrInvalidParameter)
+	}
+	biter, ok := wrapper.(wrapping.KeyBytes)
+	if !ok {
+		return nil, fmt.Errorf("%s: wrapper does not implement required KeyBytes interface: %w", op, wrapping.ErrInvalidParameter)
+	}
+	b, err := biter.GetKeyBytes()
+	if err != nil {
+		return nil, fmt.Errorf("%s: unable to get current key bytes: %w", op, err)
+	}
+	if b == nil {
+		return nil, fmt.Errorf("%s: wrapper missing bytes: %w", op, wrapping.ErrInvalidParameter)
+	}
+	reader := hkdf.New(sha256.New, b, salt, info)
+	return &io.LimitedReader{
+		R: reader,
+		N: lenLimit,
+	}, nil
+}
--- a/crypto/hmac_sha256.go
+++ b/crypto/hmac_sha256.go
@@ -1,0 +1,97 @@
+package crypto
+
+import (
+	"context"
+	"crypto/ed25519"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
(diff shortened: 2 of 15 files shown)


errors.go
```diff
@@ -0,0 +1,8 @@
+package wrapping
+
+import (
+	"errors"
+)
+
+// ErrInvalidParameter represents an invalid parameter error
+var ErrInvalidParameter = errors.New("invalid parameter")
```
jefferai, line 8: Can you either put this in const.go with the other error or can you move that one in here too?

wrapper.go
```diff
@@ -43,3 +43,10 @@ type Wrapper interface {
 	// the underlying wrapper. Supported options: WithAad.
 	Decrypt(ctx context.Context, ciphertext *BlobInfo, options ...Option) ([]byte, error)
 }
+
+// KeyBytes defines an optional interface for wrappers to implement that returns
+// the "current" key bytes
+type KeyBytes interface {
```
jefferai, line 49: I assume the idea here is that it's wrapper-specific what these bytes are? Cause in some cases it could be something like an asymmetric key mapping to big ints rather than e.g. a byte slice -- but it could be marshaled in some standardized way (pkcs<X> for instance).

crypto/derived_reader.go
```diff
@@ -0,0 +1,46 @@
...
+		op     = "reader.NewDerivedReader"
+		minLen = 20
+	)
+	if wrapper == nil {
+		return nil, fmt.Errorf("%s: missing wrapper: %w", op, wrapping.ErrInvalidParameter)
+	}
+	if lenLimit < minLen {
+		return nil, fmt.Errorf("%s: lenLimit must be >= %d: %w", op, minLen, wrapping.ErrInvalidParameter)
+	}
+	biter, ok := wrapper.(wrapping.KeyBytes)
```
jefferai, line 30: What's a biter? Kinda like a rugrat?

crypto/derived_reader.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// DerivedReader returns a reader from which keys can be read, using the
+// given wrapper, reader length limit, salt and context info. Salt and info can
+// be nil.
+//
+// Example:
+//	reader, _ := crypto.NewDerivedReader(wrapper, userId, jobId)
+// 	key := ed25519.GenerateKey(reader)
+func NewDerivedReader(wrapper wrapping.Wrapper, lenLimit int64, salt, info []byte) (*io.LimitedReader, error) {
```
jefferai, line 19: Salt and info are both optional. We could consider turning those into global options -- they're used in the aead package for derived keys too. Maybe that's _too_ global for what we need.

crypto/hmac_sha256.go
```diff
@@ -0,0 +1,97 @@
...
+
+// HmacSha256WithPrk will HmacSha256 using the provided prk.  See HmacSha256 for
+// options supported.
+func HmacSha256WithPrk(ctx context.Context, data, prk []byte, opt ...wrapping.Option) (string, error) {
+	opt = append(opt, WithPrk(prk))
+	return HmacSha256(ctx, data, nil, nil, nil, opt...)
+}
+
+// HmacSha256 the provided data. Supports WithPrefix, WithEd25519 and WithPrk
+// options. WithEd25519 is a "legacy" way to complete this operation and should
```
jefferai, line 25: I wonder if you should just not even document WithEd25519 and put a deprecation notice on the option. Or call it something like WithDeprecatedEd25519. Arguably we shouldn't even support it at all -- it feels weird to add new code that is already deprecated.

crypto/hmac_sha256.go
```diff
@@ -0,0 +1,97 @@
...
+// not be used in new operations unless backward compatibility is needed. The
+// WithPrefix option will prepend the prefix to the hmac-sha256 value.
+func HmacSha256(ctx context.Context, data []byte, cipher wrapping.Wrapper, salt, info []byte, opt ...wrapping.Option) (string, error) {
```
jefferai, line 28: Salt and info are optional

jefferai:
As far as package naming -- I wonder if there should be one package called something like "derived" or "derivations" and it could contain the NewDerivedReader and possibly the `aead.NewDerivedWrapper` could also be folded into that package and made more general using your KeyBytes interface.

Then a second package could be something like `hashing` or `hmacing` or so?

crypto/hmac_sha256.go
jimlambrt: Well... Boundary uses it, so if we don't add it then boundary won't be able to use this.

jimlambrt: happy to mark it deprecated though.

crypto/derived_reader.go
jimlambrt: converted to options within this pkg.

crypto/hmac_sha256.go
jimlambrt: ty!

crypto/hmac_sha256.go
jimlambrt: 🤔 about this a bit more... will all ed25519 options be deprecated going forward?  I think, it's just this particular op that it's deprecated.

jefferai (approved):

PR #66, opened 2022-03-19
feature: begin building out extras/kms module

jimlambrt:
This PR includes a set of commits which define the new extras/kms module along with:
- basic workflows via GitHub actions and a Makefile
- the database schema
- some test fixtures
- NewRepository(...) and Repository.ValidateVersion()
- the beginnings of a README and package docs. 

The plan is to build the new module via PRs into the `jimlambrt-extras-kms` long lived branch.  Once the feature is complete, we'll PR/merge that long-lived branch to main. 

Reviewers: the feature commits in this PR are based on work within Boundary's kms repository. 

@mgaffney @tmessi : I'd appreciate some careful review of the schemas by one of you.  ty!


extras/kms/option.go
```diff
@@ -0,0 +1,69 @@
...
+// WithLimit provides an option to provide a limit. Intentionally allowing
+// negative integers. If WithLimit < 0, then unlimited results are returned. If
+// WithLimit == 0, then default limits are used for results.
+func WithLimit(limit int) Option {
+	return func(o *options) {
+		o.withLimit = limit
+	}
+}
+
+// WithRootWrapper sets the external root wrapper for a given scope
```
johanbrandhorst, line 39: Mega nit: some of these comments are lacking a full stop 😬.

extras/kms/option.go
```diff
@@ -0,0 +1,69 @@
...
+func WithKeyId(keyId string) Option {
+	return func(o *options) {
+		o.withKeyId = keyId
+	}
+}
+
+// WithOrderByVersion provides an option to specify ordering by the
+// CreateTime field.
+func WithOrderByVersion(orderBy OrderBy) Option {
+	const col = "version"
```
johanbrandhorst, line 65: This appears unused?

extras/kms/testing.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/golang-migrate/migrate/v4/database"
+	"github.com/golang-migrate/migrate/v4/database/postgres"
+	"github.com/golang-migrate/migrate/v4/database/sqlite"
+	source "github.com/golang-migrate/migrate/v4/source"
+	"github.com/golang-migrate/migrate/v4/source/httpfs"
+	pgDriver "gorm.io/driver/postgres"
+
+	"github.com/hashicorp/go-dbw"
+	"github.com/hashicorp/go-kms-wrapping/extras/kms/v2/migrations"
+	"github.com/stretchr/testify/require"
```
johanbrandhorst, line 22: Nit: this import grouping appears inconsistent with other files.

.github/workflows/go.yml
```diff
@@ -0,0 +1,78 @@
...
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  # Label of the container job
+  sqlite:
+    strategy:
+      matrix:
+        go: ["1.17", "1.16"]
```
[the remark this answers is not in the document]
jimlambrt: done. ty!

.github/workflows/go.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Tests
+        working-directory: ./extras/kms
+        run: make test-sqlite
+
+  postgres:
+    strategy:
+      matrix:
+        dbversion: ["postgres:latest"]
+        go: ["1.17", "1.16"]
```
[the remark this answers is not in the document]
jimlambrt: done. ty!

.github/workflows/go.yml
```diff
@@ -0,0 +1,78 @@
...
+
+    services:
+      postgres:
+        image: ${{ matrix.dbversion }}
+        env:
+          POSTGRES_PASSWORD: go_db
+          POSTGRES_USER: go_db
+          POSTGRES_DB: go_db
+        ports:
+          - 9920:5432
```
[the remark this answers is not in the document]
jimlambrt: since the job uses the Makefile, they sort of need to match.  I mapped them in the Makefile because I'm never a fan of something that assumes it can just use the std port for a service for its unit tests.

extras/kms/option.go
jimlambrt: there are a few options pulled in from boundary which aren't used yet, but will be in subsequent PRs.  didn't feel like it was a big deal to leave them as long as they are used before we merge to main.

extras/kms/Makefile
```diff
@@ -0,0 +1,27 @@
+
+
```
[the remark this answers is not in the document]
jimlambrt: honestly, I'm not a dev that's bothered by extra while space.  If there's a std formatter available and we can enforce style like this via tooling then I'm all for it... otherwise I sort of personally gave up decades ago worry about bits like this.... now if there are formatters like `fmt`, `gofumpt`, `goimports` which a team chooses to align on, which we can automate, then of course I'm very happy to oblige.

extras/kms/migrations/sqlite/02_version.up.sql
```diff
@@ -0,0 +1,34 @@
+
+
```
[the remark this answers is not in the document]
jimlambrt: chuckle.

extras/kms/migrations/postgres/01_domain_types.up.sql
```diff
@@ -0,0 +1,101 @@
...
+    return old;
+  end if;
+end;
+$$ language plpgsql;
+comment on function
+  update_time_column()
+is
+  'function used in before update triggers to properly set update_time columns';
+
+commit;
```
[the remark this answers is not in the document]
jimlambrt: not sure go-migrate cares..

extras/kms/testing.go
jimlambrt: ty!

jimlambrt: not sure why the language server didn't fix that in vscode.

extras/kms/migrations/postgres/01_domain_types.up.sql
```diff
@@ -0,0 +1,101 @@
+begin;
+
+create domain wt_private_id as text
```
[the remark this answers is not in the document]
jimlambrt: `wt_` is leagacy and aligns with boundary's naming convention but I'm thinking now it's sort of silly to follow that here.  ty!

extras/kms/migrations/postgres/03_keys.up.sql
```diff
@@ -0,0 +1,45 @@
+begin;
+
+create or replace function
+  kms_root_key_version_column()
+  returns trigger
+as $$
+begin
+  update kms_root_key_version set version =
+  (
+    select max(coalesce(version,1)) + 1
```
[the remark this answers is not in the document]
jimlambrt: Agreed about the it being 2 vs 1, I think I've fixed that.

Not sure about omitting max though.

.github/workflows/go.yml
```diff
@@ -0,0 +1,78 @@
...
+          key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('tests/go.mod') }}
+
+      - name: Tests
+        working-directory: ./extras/kms
+        run: make test-sqlite
+
+  postgres:
+    strategy:
+      matrix:
+        dbversion: ["postgres:latest"]
```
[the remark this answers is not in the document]
jimlambrt: ty!

extras/kms/option.go
johanbrandhorst: I didn't mean that the function is unused, I mean that this constant appears to be unused.

jimlambrt: oh!  ty!

.github/workflows/go.yml
```diff
@@ -0,0 +1,78 @@
+name: Go
+
+on:
+  push:
+    branches: [main, "jimlambrt-extras-kms*"] # jimlambrt-extras-kms* is tmp until this initial work is done.
+  pull_request:
+    branches: [main, "jimlambrt-extras-kms*"] # jimlambrt-extras-kms* is tmp until this initial work is done.
```
johanbrandhorst, line 7: Do we want to remove this after merge or before?

extras/kms/Makefile
```diff
@@ -0,0 +1,25 @@
+PHONY: tools
+tools:
+	go generate -tags tools tools/tools.go
+
+.PHONY: fmt
+fmt:
+	gofumpt -w $$(find . -name '*.go')
```
johanbrandhorst, line 7: Minor note, I've been using this:
```shell
$ grep -L -R "^\/\/ Code generated .* DO NOT EDIT\.$" --include="*.go" | xargs -n 1 gofumpt -w
```
Which ensure that generated files that conform to https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source do not get formatted (like protobuf files). Might be a bit overkill in this repo but wanted to mention it 😁.

johanbrandhorst (approved):

tmessi (commented):
I am still doing a detailed comparison of the schema, but before I get too deep into it, I want to confirm that I am following the approach here. In boundary there are [explicitly named tables for different keys](https://github.com/hashicorp/boundary/blob/main/internal/db/schema/migrations/oss/postgres/0/31_keys.up.sql), i.e. `kms_database_key`, `kms_token_key`, but what I see here is a `kms_data_key` with a `purpose` column. So the idea is that `purpose` could be something like `database` or `token`?

I also have some questions/comments about how we want to deal with conflicting names or common functions between this library and boundary. In my opinion that library should be stand-alone and not conflict with boundary, even if that means the same function is implemented twice but with different names, but I would like to hear your and others thoughts on this.

extras/kms/data_key_version.go
```diff
@@ -0,0 +1,111 @@
...
+
+	k := &DataKeyVersion{
+		DataKeyId:        dataKeyId,
+		RootKeyVersionId: rootKeyVersionId,
+		Key:              key,
+	}
+	return k, nil
+}
+
+// Clone creates a clone of the OidcKeyVersion
```
johanbrandhorst, line 54: 😬 
```suggestion
// Clone creates a clone of the DataKeyVersion
```

extras/kms/root_key.go
```diff
@@ -0,0 +1,39 @@
...
+
+// TableName returns the tablename
+func (k *RootKey) TableName() string { return "kms_root_key" }
+
+// Clone creates a clone of the RootKeyVersion
+func (k *RootKey) Clone() *RootKey {
+	return &RootKey{
+		PrivateId:  k.PrivateId,
+		ScopeId:    k.ScopeId,
+		CreateTime: k.CreateTime}
```
johanbrandhorst, line 38: Nit: is this intentional?
```suggestion
		CreateTime: k.CreateTime,
	}
```

extras/kms/root_key_version.go
```diff
@@ -0,0 +1,104 @@
...
+	CtKey []byte `json:"ct_key,omitempty" gorm:"column:key;not_null" wrapping:"ct,key_data"`
+	// version of the key data.  This is not used for optimistic locking, since
+	// key versions are immutable.  It's just the version of the key.
+	// @inject_tag: `gorm:"default:null"`
+	Version uint32 `json:"version,omitempty" gorm:"default:null"`
+	// CreateTime from the db
+	CreateTime time.Time `json:"create_time,omitempty" gorm:"default:current_timestamp"`
+}
+
+// NewRootKey creates a new in memory root key. No optionsare currently
```
johanbrandhorst, line 32:
```suggestion
// NewRootKey creates a new in memory root key. No options are currently
```

johanbrandhorst (approved):

extras/kms/migrations/postgres/04_keys.up.sql
```diff
@@ -0,0 +1,114 @@
+begin;
+
+create table kms_root_key (
+  private_id kms_private_id primary key,
+  scope_id kms_scope_id not null unique, -- there can only be one root key for a scope.
```
tmessi, line 5: in boundary this is a foreign key with `on delete cascade`, how would that relationship be managed with this module?

extras/kms/migrations/postgres/02_version.up.sql
```diff
@@ -0,0 +1,31 @@
...
+before
+insert on kms_schema_version
+  for each row execute procedure kms_default_create_time();
+
+create trigger update_time_column 
+before 
+update on kms_schema_version 
+	for each row execute procedure update_time_column();
+
+insert into kms_schema_version(version) values('v0.0.1');
```
tmessi, line 29: So this version will be tied to a tag? I would expect this to correspond to the latest migration number, just curious why the use of this string?

tmessi (commented):
I have a couple questions, but otherwise looks good to me.

extras/kms/root_key_version.go
jimlambrt: ty!

extras/kms/migrations/postgres/04_keys.up.sql
jimlambrt: I need to document this in the README.  I'm thinking that users will likely need to add a foreign key constraint that prevents orphans, but since we have no way of determining the table name for scopes, it will be left to the user to add the constraint with guidance from our documentation.  Happy to hear other suggestions.

extras/kms/migrations/postgres/02_version.up.sql
jimlambrt: A future PR will include a runtime check of this against the tagged module version.  We can't use the "standard" migration numbers since the migrations from this module may likely need to be integrated with an application's existing migrations which will likely have different migration sequences/timestamps.

.github/workflows/go.yml
```diff
@@ -0,0 +1,78 @@
...
+          key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('tests/go.mod') }}
+
+      - name: Tests
+        working-directory: ./extras/kms
+        run: make test-sqlite
+
+  postgres:
+    strategy:
+      matrix:
+        dbversion: ["postgres:14.2", "postgres:13.6"]
```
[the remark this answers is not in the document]
jimlambrt: I've updated this to test back to Postgres 10.  ty!

extras/kms/migrations/postgres/04_keys.up.sql
tmessi: The only thought that is coming to mind (and I havnt thought it through at all) would be to introduce some kind of templating.

tmessi (approved):

PR #70, opened 2022-04-05
Explicitly provide the KeyID to Decrypt operations

sgmiller:
(But with an env var to override this behavior in a pinch).  This addresses
seal wrap migration problems customers are having where Vault is not able to
detect that a rewrap needs to start because the AWS wrapper is able to
decrypt the previous value with the "new" key because in fact it's using the
old key via the implicit ciphertext metadata.

### DIFF
--- a/wrappers/awskms/awskms.go
+++ b/wrappers/awskms/awskms.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"sync/atomic"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -21,6 +22,7 @@ import (
 const (
 	EnvAwsKmsWrapperKeyId   = "AWSKMS_WRAPPER_KEY_ID"
 	EnvVaultAwsKmsSealKeyId = "VAULT_AWSKMS_SEAL_KEY_ID"
+	EnvAWSKMSDecryptWithImplicitKeyID = "AWSKMS_DECRYPT_WITH_IMPLICIT_KEY_ID"
 )
 
 const (
@@ -46,6 +48,7 @@ type Wrapper struct {
 	roleSessionName      string
 	webIdentityTokenFile string
 	keyNotRequired       bool
+	decryptWithImplicitKeyID bool
 
 	currentKeyId *atomic.Value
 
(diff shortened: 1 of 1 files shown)


wrappers/awskms/awskms.go
```diff
@@ -97,6 +100,16 @@ func (k *Wrapper) SetConfig(_ context.Context, opt ...wrapping.Option) (*wrappin
 		return nil, fmt.Errorf("key id not found in env or config for aws kms wrapper configuration")
 	}
 
+	k.currentKeyId.Store(k.keyId)
```
stevendpclark, line 103: Should this not be wrapped in an if k.keyId != nil ? I believe the Store will panic if k.keyNotRequired is set?

schultz-is (approved):

wrappers/awskms/awskms.go
```diff
@@ -46,6 +48,7 @@ type Wrapper struct {
 	roleSessionName      string
 	webIdentityTokenFile string
 	keyNotRequired       bool
+	decryptWithImplicitKeyID bool
```
chrishoffman, line 51: go fmt

wrappers/awskms/awskms.go
stevendpclark: nvm keyId will be a blank not a nil...

stevendpclark (approved):

wrappers/awskms/awskms.go
```diff
@@ -244,10 +263,9 @@ func (k *Wrapper) Decrypt(_ context.Context, in *wrapping.BlobInfo, opt ...wrapp
 		plaintext = output.Plaintext
 
 	case AwsKmsEnvelopeAesGcmEncrypt:
-		// KeyId is not passed to this call because AWS handles this
-		// internally based on the metadata stored with the encrypted data
 		input := &kms.DecryptInput{
 			CiphertextBlob: in.KeyInfo.WrappedKey,
+			KeyId: keyID,
```
chrishoffman, line 268: I'm not sure this is the correct key to pass to the decrypt operation. This could be a key alias or a fully qualified ARN. In the case of key alias, the key could have been rotated underneath that no longer matches the key used to encrypt the value. I think you'll want to use the in.KeyInfo.KeyID instead. I believe this will only work for wrapped keys though.

PR #75, opened 2022-04-10
refactor (extras/kms): reduce the API surface to only what's needed

jimlambrt:


extras/kms/kms.go
```diff
@@ -204,6 +216,132 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
+func (k *Kms) ValidateSchema(ctx context.Context) (string, error) {
+	const op = "kms.(Repository).validateVersion"
+	return k.repo.ValidateSchema(ctx)
+}
+
+// ReconcileKeys will reconcile the keys in the kms against known possible
+// issues.  This function reconciles the global scope unless the
+// WithScopeIds(...) option is provided
+
+// The WithRandomReadear(...) option is supported as well.  If not optional
```
johanbrandhorst, line 297:
```suggestion
// The WithRandomReadear(...) option is supported as well.  If no optional
```

extras/kms/option.go
```diff
@@ -67,34 +65,53 @@ func WithKeyId(keyId string) Option {
...
+// allow the caller to create the scope and all of its keys in the same
+// transaction.
+func WithTx(tx *dbw.RW) Option {
+	return func(o *options) {
+		o.withTx = tx
+	}
+}
+
+// WithRandomReadear(...) option allows an optional random reader to be
+// provided.  By default the the reader from crypto/rand will be used.
```
johanbrandhorst, line 112:
```suggestion
// provided.  By default the reader from crypto/rand will be used.
```

extras/kms/kms.go
```diff
@@ -204,6 +216,132 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
+	const op = "kms.(Repository).validateVersion"
+	return k.repo.ValidateSchema(ctx)
+}
+
+// ReconcileKeys will reconcile the keys in the kms against known possible
+// issues.  This function reconciles the global scope unless the
+// WithScopeIds(...) option is provided
+
+// The WithRandomReadear(...) option is supported as well.  If an optional
+// random reader is not provided (is nill), then the reader from crypto/rand
```
johanbrandhorst, line 298:
```suggestion
// random reader is not provided (is nil), then the reader from crypto/rand
```

extras/kms/kms.go
```diff
@@ -204,6 +216,132 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
+
+	if isNil(opts.withRandomReader) {
+		opts.withRandomReader = rand.Reader
+	}
+
+	var tx *dbw.RW
+	switch {
+	case opts.withTx != nil:
+		if ok := opts.withTx.IsTx(); !ok {
+			return fmt.Errorf("%s: provided transaction has no inflight tranaction: %w", op, ErrInvalidParameter)
```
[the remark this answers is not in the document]
jimlambrt: ty!

extras/kms/kms.go
```diff
@@ -204,6 +216,132 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
+		tx = opts.withTx
+	default:
+		tx, err = k.repo.writer.Begin(ctx)
+		if err != nil {
+			return fmt.Errorf("%s: unable to begin transaction: %w", op, err)
+		}
+	}
+
+	if _, err := createKeysTx(ctx, tx, rootWrapper, opts.withRandomReader, scopeId, purposes...); err != nil {
+		rollBackErr := tx.Rollback(ctx)
```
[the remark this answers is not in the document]
jimlambrt: I agree that we should do this rollback if a tx is provided.  Ty!

extras/kms/kms.go
```diff
@@ -204,6 +216,155 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
+		}
+		r = opts.withReader
+		w = opts.withWriter
+	default:
+		localTx, err = k.repo.writer.Begin(ctx)
+		if err != nil {
+			return fmt.Errorf("%s: unable to begin transaction: %w", op, err)
+		}
+		r = localTx
+		w = localTx
```
johanbrandhorst, line 288: Instead of creating a `localTx` here and then checking for it, what do you think about deferring the operations we want to run only for local transactions?
```suggestion
		localTx, err = k.repo.writer.Begin(ctx)
		if err != nil {
			return fmt.Errorf("%s: unable to begin transaction: %w", op, err)
		}
		defer func() {
			if retErr == nil {
				if err  := localTx.Commit(ctx); err != nil {
					retErr = fmt.Errorf("%s: unable to commit transaction: %w", op, err)
				}
			} else {
				if err := localTx.Rollback(ctx); err != nil {
					retErr = multierror.Append(retErr, err)
				}
			}
		}()
		r = localTx
		w = localTx
```

Then we can remove the `localTx` checks and operations below. It would require us to use a named returned (name `retErr` here) to allow us to mutate the error in the defer.

johanbrandhorst (approved):

extras/kms/kms.go
jimlambrt: ty for the suggestion... played around a bit and decided to leave it basically as-is for now.

extras/kms/option.go
```diff
@@ -67,34 +67,65 @@ func WithKeyId(keyId string) Option {
...
+// for managing the transaction. The purpose of the WithTx(...) option is to
+// allow the caller to create the scope and all of its keys in the same
+// transaction.
+func WithTx(tx *dbw.RW) Option {
+	return func(o *options) {
+		o.withTx = tx
+	}
+}
+
+// WithRandomReadear(...) option allows an optional random reader to be
```
tmessi, line 113: typo: `s/Readear/Reader/`

tmessi (approved):
nice!

extras/kms/option.go
jimlambrt: ty!

PR #77, opened 2022-04-13
feature (extras/kms): add support to manage the cache of wrappers

jimlambrt:


johanbrandhorst (approved):
Nice!

extras/kms/kms.go
```diff
@@ -40,15 +40,20 @@ type Kms struct {
...
 // New takes in a reader, writer and a list of key purposes it will support.
 // Every kms will support a KeyPurposeRootKey by default and it doesn't need to
-// be passed in as one of the supported purposes.  No options are currently
-// supported.
-func New(r dbw.Reader, w dbw.Writer, purposes []KeyPurpose, _ ...Option) (*Kms, error) {
+// be passed in as one of the supported purposes.
+//
+// The WithCache(bool) option is supported and if enabled then a cache of the
+// wrappers is enabled.  Once enabled, the cache will speed up calls to
+// GetWrapper(...) but the caller also becomes responsible to reset/clear the
```
tmessi, line 54: how does a scope get deleted from the database?

extras/kms/migrations/postgres/04_keys.up.sql
```diff
@@ -56,7 +56,7 @@ create table kms_data_key (
     references kms_root_key(private_id)
     on delete cascade
     on update cascade,
-  purpose text not null unique
+  purpose text not null
```
tmessi, line 59: Why is this change needed here?

extras/kms/kms.go
jimlambrt: that's outside the "scope" of this package.

extras/kms/migrations/postgres/04_keys.up.sql
jimlambrt: in the course of testing the new cache bits I attempted to add a `database` DEK to two scopes and it failed which is a defect.

tmessi: Ah, I see, a few lines below we have `unique (root_key_id, purpose) -- there can only be one dek for a specific purpose per root key` which is the correct constraint here.

tmessi (approved):

PR #80, opened 2022-04-17
refactor (extras/kms): remove withRepository(...) option

jimlambrt:
It wasn't reachable from external callers anywho, and if you need
a specific reader/writer when using GetWrapper(...) then you just
need to create a new Kms with whatever reader/writer is needed.


jimlambrt:
It's a pretty straightforward change... just looking for one review before merging.

jefferai (approved):

PR #81, opened 2022-04-19
feature (extras/kms): add TestDeleteKeyPurpose, TestKmsDeleteAllKeys

jimlambrt:
These two new test functions allow callers to either delete a specific
key purpose or all the keys for testing.

### DIFF
--- a/extras/kms/testing.go
+++ b/extras/kms/testing.go
@@ -116,6 +116,16 @@ func TestDb(t *testing.T) (*dbw.DB, string) {
 	return dbw.TestSetup(t, dbw.WithTestMigrationUsingDB(testMigrationFn(t)))
 }
 
+// TestDeleteKeyPurpose allows you to delete a KeyPurpose for testing.
+func TestDeleteKeyPurpose(t *testing.T, conn *dbw.DB, purpose KeyPurpose) {
+	testDeleteWhere(t, conn, func() interface{} { i := dataKey{}; return &i }(), fmt.Sprintf("purpose='%s'", purpose))
+}
+
+// TestKmsDeleteAllKeys allows you to delete ALL the keys for testing.
+func TestKmsDeleteAllKeys(t *testing.T, conn *dbw.DB) {
+	testDeleteWhere(t, conn, func() interface{} { i := rootKey{}; return &i }(), "1=1")
+}
+
 func testMigrationFn(t *testing.T) func(ctx context.Context, db *sql.DB) error {
 	return func(ctx context.Context, db *sql.DB) error {
 		t.Helper()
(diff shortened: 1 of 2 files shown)


extras/kms/testing.go
```diff
@@ -116,6 +116,16 @@ func TestDb(t *testing.T) (*dbw.DB, string) {
 	return dbw.TestSetup(t, dbw.WithTestMigrationUsingDB(testMigrationFn(t)))
 }
 
+// TestDeleteKeyPurpose allows you to delete a KeyPurpose for testing.
```
tmessi, line 119: Does this delete a KeyPurpose, or a Key *with* the given KeyPurpose?

extras/kms/testing_test.go
```diff
@@ -16,3 +18,58 @@ func Test_stub(t *testing.T) {
...
+	require.NoError(err)
+	require.NoError(kmsCache.AddExternalWrapper(ctx, KeyPurposeRootKey, extWrapper))
+	// Make the global scope base keys
+	err = kmsCache.CreateKeys(ctx, globalScope, []KeyPurpose{databaseKeyPurpose})
+	require.NoError(err)
+
+	_, err = kmsCache.GetWrapper(ctx, globalScope, databaseKeyPurpose)
+	require.NoError(err)
+
+	TestDeleteKeyPurpose(t, db, databaseKeyPurpose)
```
tmessi, line 47: can this test be done in a way that confirms that this *only* deleted what it was supposed to, and did not just delete all of the keys?

extras/kms/testing_test.go
```diff
@@ -16,3 +18,58 @@ func Test_stub(t *testing.T) {
 	t.Log("rows: ", rows)
 	require.NoError(t, err)
 }
+
+func Test_TestDeleteKeyPurpose_TestDeleteAllKeys(t *testing.T) {
```
tmessi, line 22: Is there a reason to combine these as a single test instead of having two tests? I would find it easier to see the setup, execution, and assertions for each test function if they were separate tests.

extras/kms/testing.go
jimlambrt: It deletes all the keys for a given purpose.  I'll try to make that more clear.

extras/kms/testing_test.go
jimlambrt: most of it is setup and it's like an additional 10 lines to test TestDeleteAllKeys after that setup.. just didn't make sense to me to split them.

extras/kms/testing_test.go
jimlambrt: yep. ty!

extras/kms/testing_test.go
tmessi: The extra comments help :+1:

tmessi (approved):

PR #86, opened 2022-05-02
tests (envelope): add some additional tests for the envelope

jimlambrt:


jefferai (approved):
LGTM! Can you add that warning to EnvelopeEncrypt too?

PR #90, opened 2022-05-14
refactor (extras/kms): remove direct sqlite dependency

jimlambrt:


.github/workflows/go.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('tests/go.mod') }}
 
       - name: Build
-        run: find . -name go.mod -execdir go build ./... \;
+        run: CGO_ENABLED=0 find . -name go.mod -execdir go build ./... \;
```
johanbrandhorst, line 33: Should we use the `env` option instead of setting it like this?
```suggestion
        env:
          CGO_ENABLED: 0
        run: find . -name go.mod -execdir go build ./... \;
```

johanbrandhorst (approved):

jefferai (approved):

.github/workflows/go.yml
jimlambrt: The GitHub `env` option is incredibly useful if you need to setup one (or many) env vars which are used across more than one step.    In this particular case, it's only being used in one instance a plain old inline env var is more readable.

tmessi: The `env` option can also be set for specific steps, like the suggested change @johanbrandhorst provided. I personally find the `env` option more readable even in this case. I think even with a single command there is a point where too many inline env vars is less readable by just about anyone's standard, so I saw just always just the `env` option so there is no need to agree on how many is too many.

johanbrandhorst: I also realized that my example is probably broken (yay yaml!), we might need to quote the value so it doesn't get interpreted as a number or boolean (??!?).

PR #91, opened 2022-05-20
feature (extras/kms): add rotation of keys (with optional rewrap)

jimlambrt:


extras/kms/kms.go
```diff
@@ -289,55 +289,169 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+		}
+	}
+
+	return nil
+}
+
+// RotateKeys will rotate the root key versions and all DEKs for the current
+// KeyPurpose(s) of the KMS.
+//
+// If an optional WithRewrap(...) is request, then all keys will be re-encrypted
```
johanbrandhorst, line 317:
```suggestion
// If an optional WithRewrap(...) is requested, then all keys will be re-encrypted
```

extras/kms/kms.go
```diff
@@ -289,55 +289,169 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+// with the current root key wrapper.
+//
+// WithTx(...) and WithReaderWriter(...) options are supported which allow the
+// caller to pass an inflight transaction to be used for all database
+// operations.  If WithTx(...) or WithReaderWriter(...) are used, then the
+// caller is responsible for managing the transaction.  If neither WithTx or
+// WithReaderWriter are specified, then RotateKeys will rotated the scope's keys
+// within it's own transaction, which will be managed by RotateKeys.
```
johanbrandhorst, line 325:
```suggestion
// within its own transaction, which will be managed by RotateKeys.
```

extras/kms/kms.go
```diff
@@ -289,55 +289,169 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+//
+// The WithRandomReader(...) option is supported.  If no optional random reader
+// is provided, then the reader from crypto/rand will be used as a default.
+//
+// Options supported: WithRandomReader, WithTx, WithRewrap
```
johanbrandhorst, line 330: Given the above comments I assume `WithReaderWriter` should be here too
```suggestion
// Options supported: WithRandomReader, WithTx, WithReaderWriter, WithRewrap
```

extras/kms/kms.go
jimlambrt: ty!

extras/kms/kms.go
jimlambrt: ty!

extras/kms/kms.go
jimlambrt: ty

extras/kms/migrations/postgres/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,24 @@
+begin;
+
+
+-- we need to make the key and version columns mutable in order to support
+-- rewrapping them root key versions.
```
jefferai, line 5: I like the colloquial "rewrapping them root key versions" but I am guessing some word or so is actually missing :-)

jefferai: Or maybe you just meant "the"

extras/kms/migrations/postgres/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,24 @@
...
+
+create trigger kms_immutable_columns
+before
+update on kms_root_key_version
+  for each row execute procedure kms_immutable_columns('private_id', 'root_key_id', 'create_time');
+
+
+
+-- we need to make the key and version columns mutable in order to support
+-- rewrapping them data key version.
```
jefferai, line 16: Also here

extras/kms/migrations/sqlite/05_key_rewrap.up.sql
```diff
@@ -0,0 +1,36 @@
+begin;
+
+-- we need to make the key and version columns mutable in order to support
+-- rewrapping them root key versions.
```
jefferai, line 4: And here and below

extras/kms/repository_root_key_version.go
```diff
@@ -174,3 +176,96 @@ func (r *repository) ListRootKeyVersions(ctx context.Context, keyWrapper wrappin
 	}
 	return versions, nil
 }
+
+// rewrapRootKeyVersionsTx will rewrap (re-encrypt) the root key versions for a
+// given rootKeyId with the latest wrapper.
+// This function encapsulates all the work required within a dbw.TxHandler and
+// allows this capability to be shared with other repositories or just called
+// within a transaction.  To be clear, this repository function doesn't include
+// its own transaction and is intended to be used within a transaction provide
```
jefferai, line 185:
```suggestion
// its own transaction and is intended to be used within a transaction provided
```

jefferai: (And elsewhere this was copied and pasted)

extras/kms/repository_root_key_version.go
```diff
@@ -174,3 +176,96 @@ func (r *repository) ListRootKeyVersions(ctx context.Context, keyWrapper wrappin
...
+	if rootKeyId == "" {
+		return nil, fmt.Errorf("%s: missing root key id: %w", op, ErrInvalidParameter)
+	}
+
+	if isNil(writer) {
+		return nil, fmt.Errorf("%s: missing writer: %w", op, ErrInvalidParameter)
+	}
+	opts := getOpts(opt...)
+	if isNil(opts.withRandomReader) {
+		opts.withRandomReader = rand.Reader
```
jefferai, line 249: This seems like it should be part of defaultOpts?

extras/kms/repository_root_key_version_test.go
```diff
@@ -523,3 +526,326 @@ func TestRepository_ListRootKeyVersions(t *testing.T) {
...
+			for _, currRkv := range currentRootKeyVersions {
+				assert.NotEqual(got, currRkv)
+			}
+
+			newRootKeyVersions, err := tc.repo.ListRootKeyVersions(testCtx, tc.rootWrapper, tc.rootKeyId)
+			require.NoError(err)
+			sort.Slice(newRootKeyVersions, func(i, j int) bool {
+				return newRootKeyVersions[i].PrivateId < newRootKeyVersions[j].PrivateId
+			})
+			assert.Equal(len(newRootKeyVersions), len(currentRootKeyVersions)*2)
```
jefferai, line 848: I think one thing that might be good to add here is after rotation ensuring that encrypting a value uses the new key version (and maybe is not decryptable with any of the previous key versions).

extras/kms/repository_root_key.go
```diff
@@ -155,3 +155,18 @@ func (r *repository) ListRootKeys(ctx context.Context, opt ...Option) ([]*rootKe
 	}
 	return keys, nil
 }
+
+// ScopeRootKey will lookup the rootKey for a given scope id. Supported options: WithReader
+func (r *repository) ScopeRootKey(ctx context.Context, scopeId string, opt ...Option) (*rootKey, error) {
+	const op = "kms.(repository).FindRootKey"
```
jefferai, line 161: FindRootKey vs ScopeRootKey

extras/kms/repository_data_key_version.go
```diff
@@ -165,7 +166,7 @@ func (r *repository) LatestDataKeyVersion(ctx context.Context, rkvWrapper wrappi
 }
 
 // ListDataKeyVersions will lists versions of a key. Supported options:
-// WithLimit, WithOrderByVersion
+// WithLimit, WithOrderByVersion, WithReader
 func (r *repository) ListDataKeyVersions(ctx context.Context, rkvWrapper wrapping.Wrapper, databaseKeyId string, opt ...Option) ([]*dataKeyVersion, error) {
 	const op = "kms.(repository).ListDataVersions"
```
jefferai, line 171: ListDataVersions vs ListDataKeyVersions

extras/kms/migrations/postgres/05_key_rewrap.up.sql
jimlambrt: ty

extras/kms/migrations/postgres/05_key_rewrap.up.sql
jimlambrt: ty

extras/kms/migrations/sqlite/05_key_rewrap.up.sql
jimlambrt: ty

jimlambrt: ty

extras/kms/repository_root_key_version.go
jimlambrt: ty

extras/kms/repository_data_key_version.go
```diff
@@ -186,3 +187,129 @@ func (r *repository) ListDataKeyVersions(ctx context.Context, rkvWrapper wrappin
...
+	r, err := newRepository(reader, writer)
+	if err != nil {
+		return fmt.Errorf("%s: unable to create repo: %w", op, err)
+	}
+	dataKeys, err := r.ListDataKeys(ctx, withPurpose(purpose), withRootKeyId(rootKeyId), withReader(reader))
+	switch {
+	case err != nil:
+		return fmt.Errorf("%s: unable to lookup data key for %q: %w", op, purpose, err)
+	case len(dataKeys) == 0:
+		// this is NOT an error, there's just not data key to rotate for this purpose.
```
jefferai, line 290: Wouldn't it be an error to call this function without having an actual data key? It seems like if you don't have a data key for a purpose you shouldn't be calling the rotation function for the data key versions with that purpose.

extras/kms/repository_root_key_version.go
jimlambrt: agreed.  ty!

extras/kms/kms.go
```diff
@@ -289,55 +289,169 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+		}
+	}
+
+	return nil
+}
+
+// RotateKeys will rotate the scope's root key version and all DEKs for the
+// current KeyPurpose(s) of the KMS.
+//
+// If an optional WithRewrap(...) is requested, then all keys will be re-encrypted
```
jefferai, line 317: Shouldn't we also have an ability to rewrap without requiring rotation? That is, a corresponding RewrapKeys function?

extras/kms/repository_root_key.go
jimlambrt: `Find` feels like it's something that can return multiple results and LookupRootKey(...) is already taken.  Not the best name but it seems to accurately describe the what's returned and it's not overly verbose

extras/kms/repository_data_key_version.go
jimlambrt: 🤔 not much better... might just keep it the same for now.

extras/kms/kms.go
```diff
@@ -289,55 +289,169 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+// with the current root key wrapper.
+//
+// WithTx(...) and WithReaderWriter(...) options are supported which allow the
+// caller to pass an inflight transaction to be used for all database
+// operations.  If WithTx(...) or WithReaderWriter(...) are used, then the
+// caller is responsible for managing the transaction.  If neither WithTx or
+// WithReaderWriter are specified, then RotateKeys will rotated the scope's keys
```
jefferai, line 324:
```suggestion
// WithReaderWriter are specified, then RotateKeys will rotate the scope's keys
```

extras/kms/repository_data_key_version.go
jimlambrt: maybe... I was thinking of the situation where you decide to rotate keys before reconciling them.

extras/kms/kms.go
jimlambrt: ty

extras/kms/kms.go
jimlambrt: it might be nice to have both and we can always add that in the future.   At this point, I wanted to get rotation implemented first.  perhaps I'll update the PR title to reflect that it's rotation with optional rewrap.

extras/kms/repository_data_key_version.go
jimlambrt: maybe... I was thinking about the scenario where the caller rotates the keys before reconciling them.

extras/kms/repository_root_key_version_test.go
jimlambrt: done and I added this to the data key version unit test as well.

extras/kms/repository_data_key_version.go
tmessi: I think @jefferai is calling out the mismatch between the function name and the `op` string.

extras/kms/repository_root_key.go
tmessi: Same here, I think @jefferai is calling out the mismatch between function and `op` string.

However, I agree that `Find` implies the potential for multiple results, but I do find this function name opaque. Perhaps something like `GetRootKey`, `RetriveRootKey`, `RootKeyForScope`, or `LookupRootKeyForScope`?

extras/kms/kms.go
```diff
@@ -289,55 +289,205 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+		return fmt.Errorf("%s: %w", op, err)
+	}
+	if localTx != nil {
+		if err := localTx.Commit(ctx); err != nil {
+			return fmt.Errorf("%s: unable to commit transaction: %w", op, err)
+		}
+	}
+
+	return nil
+}
```
tmessi: This seems unrelated to the addition of key rotation. Is there some context around this change? The commit that adds this also introduces the `RotateKeys` function, but its not clear how this all relates.

extras/kms/repository_data_key_version.go
jimlambrt: ty!

extras/kms/repository_root_key.go
jimlambrt: Yeah, I iterated through those names as well.   For now, I think I'll keep it the same unless there's something significantly better suggested.  ty!

tmessi: I think the issue with `ScopeRootKey` is that it is ambiguous if `Scope` is a noun or verb. To me given that functions often begin with a verb, `ScopeRootKey` seems like it is "scoping" a root key, so adding or applying a scope to a root key. I think prefer `LookupRootKeyForScope` or `LookupRootKeyByScope` since:

- The use of `ForScpoe`/`ByScope` removes the ambiguity, scope is clearly used as a noun.
- It uses the same naming as the other Lookup method, `LookupRootKey` so it will be naturally grouped with that function in docs.

extras/kms/kms.go
jimlambrt: this change is a refactor to use `k.txFromOpts(...)` this is a shared func between `CreateKeys` and `RotateKeys`

tmessi: Ah, I think I see it now. `txFromOpts` extracts functionality that used to be directly in `CreateKeys` so that it can then be used by both `CreateKeys` and `RotateKeys`.

tmessi (approved):

johanbrandhorst (approved):

extras/kms/repository_data_key_version.go
jefferai: @tmessi ding ding :-D

PR #97, opened 2022-06-06
Support multiple MSI ids for Azure Key Vault

### LINKED ISSUE #44: Feature request: Allow specifying Azure ClientID
It is possible for VMs in Azure to have multiple identities. When using both a system managed identity and a user managed identity on an instance where Vault is running, `https://github.com/hashicorp/go-kms-wrapping/blob/main/wrappers/azurekeyvault/azurekeyvault.go:getKeyVaultClient` will never fill in the ClientID returned from `newMSIConfig`. When the Authorizer function goes to pass this configuration `github.com/Azure/go-autorest/autorest:NewBearerAuthorizer`, I believe it defaults to the system-managed identity. 

In our case, the system-managed identity doesn't have access to the Azure Key Vault, it is only used for the Azure Monitor Agent.

Here is a rough trace of the relevant source.

- [Entrypoint][entrypoint]
  - [NewMSIConfig][NewMSIConfig]
    - [ManagedIdentityOptions][ManagedIdentityOptions]
  - [Authorizer][Authorizer]
    - [ServicePrincipalToken][ServicePrincipalToken]
      - [NewServicePrincipalTokenFromManagedIdentity][NewServicePrincipalTokenFromManagedIdentity]
        - [newServicePrincipalTokenFromMSI][newServicePrincipalTokenFromMSI]

[entrypoint]: https://github.com/hashicorp/go-kms-wrapping/blob/main/wrappers/azurekeyvault/azurekeyvault.go#L318
[NewMSIConfig]: https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go#L541
[Authorizer]: https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go#L750
[ServicePrincipalToken]: https://github.com/Azure/go-autorest/blob/master/autorest/azure/auth/auth.go#L739
[ManagedIdentityOptions]: https://github.com/Azure/go-autorest/blob/4e2e1bb52ffbf0cf2dc47d548e7621d68fde032d/autorest/adal/token.go#L771
[NewServicePrincipalTokenFromManagedIdentity]: https://github.com/Azure/go-autorest/blob/4e2e1bb52ffbf0cf2dc47d548e7621d68fde032d/autorest/adal/token.go#L787
[newServicePrincipalTokenFromMSI]: https://github.com/Azure/go-autorest/blob/4e2e1bb52ffbf0cf2dc47d548e7621d68fde032d/autorest/adal/token.go#L794

skatsaounis:
Fixes #44

### DIFF
--- a/wrappers/azurekeyvault/azurekeyvault.go
+++ b/wrappers/azurekeyvault/azurekeyvault.go
@@ -309,6 +309,9 @@ func (v *Wrapper) getKeyVaultClient() (*keyvault.BaseClient, error) {
 	default:
 		config := auth.NewMSIConfig()
 		config.Resource = strings.TrimSuffix(v.resource, "/")
+		if v.clientID != "" {
+			config.ClientID = v.clientID
+		}
 		authorizer, err = config.Authorizer()
 		if err != nil {
 			return nil, err


skatsaounis:
This is what is done by go-autorest itself https://github.com/Azure/go-autorest/blob/main/autorest/azure/auth/auth.go#L195 when it is requested to create an authoriser based on the given environment during step 4 MSI: https://github.com/Azure/go-autorest/blob/main/autorest/azure/auth/auth.go#L78

skatsaounis:
@jimlambrt @jefferai since you two appear in the most if not all of the commits here, could you please take a look on this PR?

jefferai:
@skatsaounis I know nothing about this code. It seems reasonable to me, however.

jefferai:
Sending a review request to @cipherboy just to ensure that Vault doesn't think this might result in a regression.

skatsaounis:
Hi @cipherboy. Did you have the time to take a look here?

cipherboy (approved):
Oh interesting, I somehow missed notifications on this one.

Yes, this makes sense to me. It appears to be pass through correctly to callers of SetConfig, so I think we should largely be in good shape here.

jefferai:
Sounds good. I'll merge and tag.

jefferai:
Done. Thanks @skatsaounis !

skatsaounis:
@jefferai good news! As soon as there is a tag ready please let me know to create a PR in Vault with a bump to go.mod. Most probably you need to create 0.7.2 as Vault is using 0.7.0 at the moment (https://github.com/hashicorp/vault/blob/ed52771d63e28f110b03c4ec9da8085fb65cba38/go.mod#L69)

cipherboy:
Hey @skatsaounis -- at some point, Vault plans to upgrade to v2 of g-k-w, but that's a much larger effort than just bumping `go.mod`.  Are you wanting this in v1.11 as well? If so, we'll need to backport that to a separate branch.

skatsaounis:
I would like to have it to Vault v1.11. The reason is that the original bug report in Vault which I mentioned above was closed in 2019 without fixing the issue.

Now regarding the issue itself. There is a scenario (Consul backend) when inside a VM in Azure there are two separate MSI ids. One for accessing the Azure Key Vault used as the seal wrapping mechanism and another one used by Consul agent for cloud auto-join. This scenario cannot work as the client_id being set in seal "azurekeyvault" block is ignored.

At the moment, the usage of at least one Azure Service Principal for any of the two functionalities is required. When this patch is consumed by Vault, no SPN will be required, with an immediate boost to security and operationalisation.

PR #99, opened 2022-06-24
refactor (extras/kms): add version to the entire kms collection and always enable the cache

jimlambrt:
Add support for versioning the entire "collection" of keys represented
by the kms.

Any time one of the keys in the collection changes, the
version will be incremented.

Adding this capability will make it easy to invalidate the kms cache.
This is especially important when the cache is shared (distributed)
across multiple processes.

This change requires incrementing the migrations version.

Given the enhancements with how the cache is handled, we're going
to always enable it, so we need to remove it as an option

### DIFF
--- a/extras/kms/kms.go
+++ b/extras/kms/kms.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 	"sync"
+	"sync/atomic"
 
 	"github.com/hashicorp/go-dbw"
 	wrapping "github.com/hashicorp/go-kms-wrapping/v2"
@@ -38,6 +39,8 @@ type Kms struct {
 	scopedWrapperCache sync.Map
 
 	externalWrapperCache sync.Map
+
+	collectionVersion uint64
 
 	purposes  []KeyPurpose
 	repo      *repository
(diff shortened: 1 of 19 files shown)


jimlambrt:
Once merged, we'll open a subsequent PR to update the example CLI

extras/kms/kms.go
```diff
@@ -205,19 +195,29 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
+			if !ok {
+				return nil, fmt.Errorf("%s: scoped wrapper is not a multi.PooledWrapper: %w", op, ErrInternal)
+			}
+			if opts.withKeyId == "" {
+				return wrapper, nil
+			}
+			if keyIdWrapper := wrapper.WrapperForKeyId(opts.withKeyId); keyIdWrapper != nil {
+				return keyIdWrapper, nil
+			}
+			// Fall through to refresh our multiwrapper for this scope/purpose from the DB
```
tmessi, line 219: Is this the behavior we want if `currVersion < atomic.LoadUint64(&k.collectionVersion)`?

tmessi (commented):
Looks good, just one question

extras/kms/kms.go
tmessi: Yea, I think it might be possible if `currVersion` eventually rolls over, which admittedly shouldn't happen for a very very long time given that it we are using `bigint` for postgres, and it looks like `int` for sqlite has the same max value. But I think if it did to happen, we wouldn't want the behavior the `default` case.

jimlambrt: agreed. ty!

extras/kms/repository.go
```diff
@@ -130,6 +130,51 @@ type vetForWriter interface {
...
+	}
+
+	rowsUpdated, err := w.Exec(ctx, sql, nil)
+	if err != nil {
+		return fmt.Errorf("%s: %w", op, err)
+	}
+	if rowsUpdated != 1 {
+		return fmt.Errorf("%s: update %q rows and expected 1: %w", op, rowsUpdated, ErrInternal)
+	}
+	return nil
```
jefferai, line 150: Would it be useful to return the current key version post-update? I realize it can always be changed later, just not sure if it would be helpful elsewhere in the current code.

extras/kms/kms_test.go
```diff
@@ -1061,6 +1082,12 @@ func TestKms_ReconcileKeys(t *testing.T) {
 					}
 				}
 			}
+
+			if tt.name != "nothing-to-reconcile" {
```
jefferai, line 1086: Smells in here

jefferai (approved):
10/10, would globally version again

extras/kms/repository.go
jimlambrt: At this point, it's not helpful.  it's not exported, so we can always change the return in the future if/when it's helpful to return the post-update version.

extras/kms/kms_test.go
jimlambrt: yep.  fixed.

PR #102, opened 2022-06-28
extras/kms package: refactor RevokeKey(...) and add ListKeys(...)

jimlambrt:


jefferai (changes requested):
I think it looks good but I have a feeling that JSON tag is a mistake

PR #108, opened 2022-09-02
Update kms Package to Reflect Key/KeyVersion Structure

DanielleMiu:
things still needing updates
- any and all tests touching the kms package
- update any references to old key version (key id particularly) to reflect new name (ie `KeyId` -> `KeyVersionId`)
  - this includes references to keys themselves (id `AddKey` -> `AddKeyVersion`)
  - `kms.addKey`
  - `kms.CreateKeys` ?
  - `kms.RevokeKey`
  - `generateKey` (from repository.go)
-  above list is not exhaustive, just what I've found

from what I can tell `repository_root_key.go` doesn't need to be changed at all and neither does `repository_data_key.go` as both the `rootKey` and `dataKey` types point to the key tables and not version tables
anything belonging to the wrapping _should not_ change


johanbrandhorst:
This opts for changing uses of Key to KeyVersion rather than introducing a new concept above Key, after all. This is a breaking change on the API, so I wanted to have a look at what the impact of that would be. I used two different code search tools to determine whether this would impact any users, [Sourcegraph](https://sourcegraph.com/search) and the [Github Codesearch tech preview](https://cs.github.com). The latter allowed me to search internal repositories[^1], while the former can confidently search the entire Github corpus[^2]. The _only_ user outside of Boundary, its mirrors and various forks, is an internal repository, and their use of the library will not be affected by these changes.

Technically I also couldn't find a stable tagged release for the `github.com/hashicorp/go-kms-wrapping/extras/kms/v2` module yet, so maybe this is OK because we're still in a pre-release state, until we tag a stable v2 version of this module?

[^1]: [Github Codesearch search results](https://cs.github.com/?scopeName=All+repos&scope=&q=NOT+repo%3Ahashicorp%2Fboundary+NOT+repo%3Ahashicorp%2Fgo-kms-wrapping+%22github.com%2Fhashicorp%2Fgo-kms-wrapping%2Fextras%2Fkms%2Fv2%22++language%3AGo)
[^2]: [Sourcegraph search results](https://sourcegraph.com/search?q=context:global+%22github.com/hashicorp/go-kms-wrapping/extras/kms/v2%22+lang:Go+NOT+repo:hashicorp/boundary+NOT+repo:hashicorp/go-kms-wrapping&patternType=standard)

johanbrandhorst:
One unfortunate consequence of this change is that we cannot change the `wrapping.Wrapper` interface, which has a `KeyId(context.Context) (string, error)` which naturally returns the data/root key version ID whenever called 😞.

extras/kms/errors.go
```diff
@@ -20,6 +20,9 @@ var (
 	// ErrKeyNotFound represents that no key was found
 	ErrKeyNotFound = errors.New("key not found")
 
+	// ErrKeyVersionNotFound represents that no key version was found
+	ErrKeyVersionNotFound = errors.New("key version not found")
```
johanbrandhorst, line 24: Not convinced about the need for this, since it is technically another breaking change on the API. Happy to remove it if we think it'll cause more problems than it'll help. Added it because we may want to be able to tell the difference between key not found and key version not found on the API.

extras/kms/examples/oidc.go
```diff
@@ -12,8 +12,8 @@ type OIDC struct {
 	CtClientSecret []byte `json:"-" gorm:"column:client_secret" wrapping:"ct,client_secret"`
 	// ClientSecret is the oidc client secret (plaintext)
 	ClientSecret string `json:"client_secret,omitempty" wrapping:"pt,client_secret"`
-	// KeyId is the wrapper key id used to encrypt/decrypt the client secret
-	KeyId string `json:"key_id,omitempty" gorm:"not_null"`
+	// KeyVersionId is the wrapper key version id used to encrypt/decrypt the client secret
+	KeyVersionId string `json:"key_id,omitempty" gorm:"not_null"`
```
johanbrandhorst, line 16: Leaning towards changing the json tag here as well to make it explicit that it is a key version that is referenced.

extras/kms/kms.go
```diff
@@ -367,47 +365,47 @@ func (k *Kms) CreateKeys(ctx context.Context, scopeId string, purposes []KeyPurp
...
+// versions before revoking them.  If it's a DEK, then you need to re-encrypt all
+// data that was encrypted with the key version before revoking it.  You must have
+// foreign key restrictions between DEK key version IDs
 // (kms_data_key_version.private_id) and columns in your tables which store the
 // wrapper key ID used for encrypt/decrypt operations, otherwise you could lose
-// access to your encrypted data when you revoke a DEK that's still being used.
-func (k *Kms) RevokeKey(ctx context.Context, keyId string) error {
-	const op = "kms.(Kms).RevokeKey"
+// access to your encrypted data when you revoke a DEK version that's still being used.
+func (k *Kms) RevokeKeyVersion(ctx context.Context, keyVersionId string) error {
```
johanbrandhorst, line 375: This is a breaking API change

extras/kms/option.go
```diff
@@ -55,11 +55,11 @@ func withLimit(limit int) Option {
 	}
 }
 
-// WithKeyId allows specifying a key ID that should be found in a scope's
+// WithKeyVersionId allows specifying a key version ID that should be found in a scope's
 // multiwrapper; if it is not found, keys will be refreshed
-func WithKeyId(keyId string) Option {
+func WithKeyVersionId(keyVersionId string) Option {
```
johanbrandhorst, line 60: This is a breaking API change

extras/kms/kms.go
```diff
@@ -241,22 +241,22 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
-		return nil, fmt.Errorf("%s: unable to find specified key ID: %w", op, ErrKeyNotFound)
+		return nil, fmt.Errorf("%s: unable to find specified key version ID: %w", op, ErrKeyVersionNotFound)
 	}
 
 	return wrapper, nil
 }
 
 // ListKeys returns the current list of kms keys
-func (k *Kms) ListKeys(ctx context.Context, scopeId string) ([]Key, error) {
+func (k *Kms) ListKeys(ctx context.Context, scopeId string) (*ScopeKeys, error) {
```
johanbrandhorst, line 259: This is a breaking API change

jimlambrt (changes requested):
Amazing work!  

I’ve suggested a few changes; primarily to limit the blast radius of the breaking changes.

extras/kms/option.go
johanbrandhorst: Great idea

extras/kms/kms.go
johanbrandhorst: Sounds good

extras/kms/errors.go
johanbrandhorst: Will remove it

extras/kms/examples/cli/main.go
```diff
@@ -123,7 +123,7 @@ func main() {
 		PrivateId:    oidcId,
 		ClientId:     "example-client-id",
 		ClientSecret: *pt,
-		KeyId:        dbWrapperKeyId,
+		KeyId:        dbWrapperKeyVersionId,
```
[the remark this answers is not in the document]
johanbrandhorst: Oops!

extras/kms/examples/cli/main.go
```diff
@@ -104,9 +104,9 @@ func main() {
 		return
 	}
 
-	// get the DEK's key id so we can save it with the oidc row (since it will
+	// get the DEK version's id so we can save it with the oidc row (since it will
 	// be used to encrypt/decrypt ciphertext in the row)
-	dbWrapperKeyId, err := dbWrapper.KeyId(mainCtx)
+	dbWrapperKeyVersionId, err := dbWrapper.KeyId(mainCtx)
```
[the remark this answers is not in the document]
johanbrandhorst: Yes, that is definitely the most unfortunate part of this change. An alternative would be to wrap the `wrapper.Wrapping` interface in our own interface (replacing KeyId() with KeyVersionId()). and update all uses within Boundary to use that interface instead. This seemed a bit too invasive and wouldn't work with other packages using the root level `wrapper.Wrapping` but I'm open to it if you think it makes sense.

extras/kms/key.go
```diff
@@ -24,12 +37,43 @@ type Key struct {
...
+		Scope:      scope,
+		CreateTime: key.CreateTime,
+		Type:       KeyTypeDek,
+		Purpose:    key.Purpose,
+	}
+}
+
+// ScopeKeys is the package containing all keys in use by a certain scope
+// including a single root key (KEK) and all data keys (DEKs)
+type ScopeKeys struct {
```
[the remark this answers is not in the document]
johanbrandhorst: Yeah, we can alternatively just return the top level fields from the function, let me change it

extras/kms/kms.go
```diff
@@ -705,7 +701,7 @@ func (k *Kms) ReconcileKeys(ctx context.Context, scopeIds []string, purposes []K
 	return nil
 }
 
-func (k *Kms) loadRoot(ctx context.Context, scopeId string, opt ...Option) (*multi.PooledWrapper, string, error) {
+func (k *Kms) loadRoot(ctx context.Context, scopeId string, opt ...Option) (_ *multi.PooledWrapper, rootKeyId string, _ error) {
```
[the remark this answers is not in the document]
johanbrandhorst: Indeed, because returning a string ID here is ambiguous, I wanted to add clarity for the caller.

extras/kms/key.go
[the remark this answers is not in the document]
johanbrandhorst: I changed ListKeys so that it is no longer a breaking change 🎉.

extras/kms/kms.go
```diff
@@ -241,69 +244,66 @@ func (k *Kms) GetWrapper(ctx context.Context, scopeId string, purpose KeyPurpose
...
-		return nil, fmt.Errorf("%s: unable to find specified key ID: %w", op, ErrKeyNotFound)
+		return nil, fmt.Errorf("%s: unable to find specified key version ID: %w", op, ErrKeyNotFound)
 	}
 
 	return wrapper, nil
 }
 
-// ListKeys returns the current list of kms keys
+// ListKeys returns the current list of kms keys.
 func (k *Kms) ListKeys(ctx context.Context, scopeId string) ([]Key, error) {
```
johanbrandhorst, line 262: I reverted this back to the old signature, which makes it less of a breaking now. Now the only break I believe is the removal of the `Version` field from the `Key` type.

extras/kms/examples/cli/main.go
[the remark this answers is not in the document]
jimlambrt: Yeah.. I agree that it's unfortunate, but it's not something we should even try to address.

jimlambrt (approved):
Amazing work!  (just one small non-blocking suggestion)

PR #110, opened 2022-09-09
Add new method for finding key users

johanbrandhorst:
The new `ListDataKeyVersionReferencers` method can be used to find tables that reference the data key version table's private_id column as a foreign key. It only works with a Postgres backend.

I couldn't find an easy way to answer this question in SQLite, so it might just be that we don't add this here since it's not portable. Thoughts?


johanbrandhorst:
This depends on https://github.com/hashicorp/go-dbw/pull/25, don't merge until that has been merged and this dependency has been updated.

jimlambrt:
Could we perhaps support sqlite for this via something like this (untested):
```
sqliteQuery = `
		select 
    		m.name
		from
    		sqlite_master m
    	join pragma_foreign_key_list(m.name) p on m.name != p."table"
		where 
			m.type = 'table'
			p."table" = 'kms_data_key_version'
			p."to" = private_id'
		`
```

johanbrandhorst:
I've removed 1.16 and added 1.19 to the test matrix. With the `go-dbw` dependency update we had a file in our dependency closure that would not build with 1.16 (specifically; https://github.com/xo/dburl/blob/v0.11.0/gen.go#L1, because it was missing the old build tag style). We could fork the dependency and fix it, but this is easier and probably should be done anyway.

extras/kms/kms_internal_test.go
```diff
@@ -1165,7 +1165,7 @@ func TestKms_RevokeRootKeyVersion(t *testing.T) {
 				require.NoError(t, err)
 
 				testInsertEncryptedData(t, dbWrapper, rw, []byte("test-plaintext"))
-				t.Cleanup(func() { testDeleteWhere(t, db, &testEncryptedData{}, "1=1", nil) })
+				t.Cleanup(func() { testDeleteWhere(t, db, &testEncryptedData{}, "1=1") })
```
johanbrandhorst, line 1168: All these little things is because of the go-dbw update, which pulls in a new (minor) version of gorm. It's impressive how many things they can break on a minor.

jimlambrt (approved):
ty!

PR #111, opened 2022-09-13
Vault-5647: fix key ID error in AWS KMS wrapper

rculpepper:
This fixes the key ID error in using the AWS KMS v2 wrapper with managed keys


sgmiller (approved):

jefferai:
@rculpepper @sgmiller Note that I've tagged a `wrappers/awskms/v2.0.1` that includes this change

PR #112, opened 2022-09-14
fix: Update DEK root key version id when rewrapping

johanbrandhorst:
If the root key version id isn't updated when rewrapping, revoking the old key version will cause the DB to attempt to drop the data key.


jimlambrt (approved):
Ty!

PR #115, opened 2022-10-28
feature (extra/kms): add WithReader(...) support to GetWrapper(...)

jimlambrt:
Allow callers to specify an optional WithReader(...) to use when calling GetWrapper(...)


johanbrandhorst (approved):
Nice!

PR #125, opened 2023-02-15
Migrate go-genproto to google-cloud-go

sgmiller:


cipherboy (approved):
Well that was easy :-)

PR #142, opened 2023-02-27
upgrade go version, all deps and gen protos with updated tool chain

jimlambrt:


.github/workflows/go.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go: ["1.19", "1.18", "1.17"]
+        go: ["1.20", "1.19", "1.18", "1.17"]
```
stevendpclark, line 13: I'm just curious, if we are saying through `go.mod` users need 1.20 and above to use this, why bother testing with older versions in CI?

jimlambrt: It's my understanding that the `go directive` `Indicates that the module was written assuming the semantics of the Go version specified by the directive.` So users of the pkg can't assume compatibility with older versions than the directive.  That doesn't mean we don't want to test it with older versions to ensure it's compatible if possible.... but we're free to deprecate older versions (and their tests) if issues arise going forward without worrying about users expecting compatibility.

stevendpclark: Alright I can get behind that definition. 

Shouldn't we then be updating all the various package's `go.mod` and not just the top level one?

jimlambrt: Yep, we will after this one is updated.  The "interior" go.mods all sort of import this one.

stevendpclark (approved):
👍

PR #143, opened 2023-02-28
chore: update additional module go versions and deps

jimlambrt:
I think, this gets most of them.   

I did have to remove 1.17 from the extras/kms test matrix.

After this is merged, I can update the few remaining example deps and we should be able to tag a release.


plugin/go.mod
```diff
@@ -3,32 +3,32 @@ module github.com/hashicorp/go-kms-wrapping/plugin/v2
 go 1.17
```
stevendpclark, line 3: I believe this should be bumped to 1.20 to match the others? I'm getting an error from a `go mod tidy` run in the plugin folder until I bump it

```
go mod tidy
github.com/hashicorp/go-kms-wrapping/plugin/v2 imports
	github.com/stretchr/testify/require imports
	github.com/stretchr/testify/assert imports
	gopkg.in/yaml.v3 tested by
	gopkg.in/yaml.v3.test imports
	gopkg.in/check.v1 loaded from gopkg.in/check.v1@v1.0.0-20190902080502-41f04d3bba15,
	but go 1.16 would select v1.0.0-20201130134442-10cb98267c6c
github.com/hashicorp/go-kms-wrapping/plugin/v2 imports
	github.com/stretchr/testify/require imports
	github.com/stretchr/testify/assert imports
	gopkg.in/yaml.v3 tested by
	gopkg.in/yaml.v3.test imports
	gopkg.in/check.v1 imports
	github.com/kr/pretty imports
	github.com/rogpeppe/go-internal/fmtsort loaded from github.com/rogpeppe/go-internal@v1.6.1,
	but go 1.16 would select v1.9.0

To upgrade to the versions selected by go 1.16:
	go mod tidy -go=1.16 && go mod tidy -go=1.17
If reproducibility with go 1.16 is not needed:
	go mod tidy -compat=1.17
For other options, see:
	https://golang.org/doc/modules/pruning
```

stevendpclark (commented):
Overall +1, I think we just need to bump the go version within `plugin/go.mod` 

Note that a `go mod tidy` removes the duplicated/older protobuf dependencies from `tencentcloudkms` and `ocikms` along with a bunch of `go.sum` cleanup which we might want to do. 

Edit: looked at the protobuf changes a bit closer (they are cleanup of the older indirect references to `github.com/golang/protobuf` and keeping `google.golang.org/protobuf`)

plugin/go.mod
jimlambrt: ty!

stevendpclark (approved):

PR #154, opened 2023-03-16
feature/refactor (extras/crypto): add Sha256SumWriter and refactor sha256sum to support WithHexEncoding

jimlambrt:
Sha256SumWriter provides multi-writer which will be used to write to a hash and produce a sum.  It implements io.Writer and optionally io.Closer

### DIFF
--- a/extras/crypto/options.go
+++ b/extras/crypto/options.go
@@ -137,6 +137,7 @@ func WithInfo(info []byte) wrapping.Option {
 	}
 }
 
+// WithHexEncoding allows an optional request to use hex encoding.
 func WithHexEncoding(withHexEncoding bool) wrapping.Option {
 	return func() interface{} {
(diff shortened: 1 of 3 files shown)


extras/crypto/options.go
```diff
@@ -135,3 +136,12 @@ func WithInfo(info []byte) wrapping.Option {
 		})
 	}
 }
+
+func WithHexEncoding(withHexEncoding bool) wrapping.Option {
```
tmessi, line 140: can you add a comment to this function?

tmessi (approved):
ty for putting this together! Just a couple minor things around some comments.

extras/crypto/options.go
jimlambrt: ty!

PR #156, opened 2023-03-23
fix(extras/crypto): Close underlying writer for Sha256SumWriter

tmessi:
A call to Close on a Sha256SumWriter would check if the io.MultiWriter
was an io.Closer and call Close on it. However, an io.MultiWriter does
not implement io.Closer, so the type assertion would never be true. Thus
if the initial io.Writer that is given to the Sha256SumWriter
implemented io.Closer, it would never be called.

This updates the Sha256SumWriter to keep a reference to the original
io.Writer, so it can perform the check against it and call the correct
Close. This allows Sha256SumWriter to be used as a transparent wrapper
of a io.WriteCloser.


jimlambrt (approved):
Ty!

PR #161, opened 2023-03-31
Add wrapper option to ignore env vars

rculpepper:
This adds a wrapper option for ignoring environment variables in wrapper config so that any wrapper can use this for configuration


schultz-is (approved):
lgtm, but i'm not super familiar with our increment by 10 protobuf field ID practice

jimlambrt:
I'm curious where this option will be supported, since I don't see any bits that check/act upon it.

cipherboy:
@jimlambrt This will be part of the Seal HA initiative.

PR #162, opened 2023-04-09
feat: classify proto type data as secret, sensitive or public

jimlambrt:


github.com.hashicorp.go.kms.wrapping.v2.types.proto
```diff
@@ -19,25 +19,31 @@ message WrapperConfig {
 // decryption in an envelope fashion
 message EnvelopeInfo {
   // Ciphertext is the ciphertext from the envelope
+  // @gotags: `class:"public"`
```
johanbrandhorst, line 22: Nit: I found that the backticks are optional, and I kinda prefer this without them? In Boundary we're not using either consistently, but I think we can do it here .
```suggestion
  // @gotags: class:"public"
```

tools/tools.go
```diff
@@ -0,0 +1,24 @@
...
+// $ make tools
+// or
+// $ go generate -tags tools tools/tools.go
+
+package tools
+
+// NOTE: This must not be indented, so to stop goimports from trying to be
+// helpful, it's separated out from the import block below. Please try to keep
+// them in the same order.
+//go:generate go install github.com/favadi/protoc-go-inject-tag
```
johanbrandhorst, line 20: Can we not just use the same way of versioning this as we do buf? I.e. just a `go install` in the Makefile?

github.com.hashicorp.go.kms.wrapping.v2.types.proto
jimlambrt: ty

Makefile
```diff
@@ -4,4 +4,13 @@ proto:
...
 	
+	# inject classification tags (see: https://github.com/hashicorp/go-eventlogger/tree/main/filters/encrypt)
+	@protoc-go-inject-tag -input=./github.com.hashicorp.go.kms.wrapping.v2.types.pb.go
+	
 .PHONY: proto
+
+
+.PHONY: tools
+tools: 
+	go install github.com/favadi/protoc-go-inject-tag
```
johanbrandhorst, line 15: Does this work? I think we should explicitly version it still.

jimlambrt: yes, it works.  happy to pin it.

.github/workflows/go.yml
```diff
@@ -161,6 +161,13 @@ jobs:
         with:
           # The 'main' branch of the GitHub repository that defines the module.
           against: "https://github.com/${GITHUB_REPOSITORY}.git#branch=main"
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v4
+        with:
+          go-version-file: go.mod    
+      - name: install go-inject-tag dependency
+        run: |
+          go install github.com/favadi/protoc-go-inject-tag@v1.4.0
```
johanbrandhorst, line 170: Shouldn't this just be `make tools`?

jimlambrt: Nope.

johanbrandhorst (approved):

PR #163, opened 2023-04-10
Vault-11623: Add support for ignoring env vars in wrappers

rculpepper:
This uses the changes in #161 to support an option to ignore env vars in wrappers used in the Vault seal


github.com.hashicorp.go.kms.wrapping.v2.types.proto
```diff
@@ -101,6 +101,8 @@ message KeyInfo {
 
 // Options holds options common to all wrappers
 message Options {
+  bool with_disallow_env_vars = 1;
+
```
schultz-is: Probably want to make this 90 to match the other occurrence.

schultz-is (approved):

PR #170, opened 2023-06-20
Adds an option to perform encryption operations with no HMAC required.

schultz-is:
Certain AES mechanisms require an HMAC when used as a seal, but this behavior is not desired when used as a managed key.


github.com.hashicorp.go.kms.wrapping.v2.types.proto
```diff
@@ -143,6 +143,9 @@ message Options {
   KeyEncoding with_wrapped_key_encoding = 80;
 
   bool with_disallow_env_vars = 90;
+
+  // WithNoHmac specifies that an HMAC is not necessary for the mechanism, even if marked as "required"
+  bool with_no_hmac = 100;
```
sgmiller, line 148: s/with_no/without?  I don't feel strongly either way.

sgmiller (approved):

github.com.hashicorp.go.kms.wrapping.v2.types.proto
schultz-is: sgtm. that's a more natural phrasing.

PR #172, opened 2023-07-12
Add new BlobInfo fields Generation and NextBlobInfo.

victorr:
Add field NextBlobInfo so that a chain of BlobInfos can be used to represent multiple encrpytions of the same value by different keys.

Add field Generation which can be used as a version of the BlobInfo chain.


sgmiller (approved):

victorr:
The motivation for this change is to make it easier for Vault to support multi seal wrapped values, especially since we want a modicum of backwards compatibility to avoid making upgrades too difficult.

Vault uses `BlobInfo` for the value of storage entries that have been seal wrapped. Vault 1.15 will introduce Seal High-Availability, which includes the ability to wrap values using multiple seals, so that values can be unsealed as long as there is at least one seal available.

As described in [VLT-246: Seal High Availability Redux](https://docs.google.com/document/d/1kEzIz0FYI--MNMgxN5HtMdHFZ5SRPDEq1gYvQaWV5wk/edit), the original plan was to introduce a new protobuf message which would have a slice of BlobInfos. However, we have decided against this approach as its implementation would require special handling not to make Vault upgrades too difficult.

The issue is that if a new Vault node were to change the value of the barrier key ring from a `BlobInfo` to that of the new type, then old Vault nodes would not be able to unseal as they would no longer be able to obtain the barrier key. While this is not strictly a problem, it makes upgrades too "brittle" as once started operators cannot easily go back to the previous version (not to mention the noise of old nodes being unable to unseal during the upgrade process).

We tried the idea of only starting to use the new type when it becomes necessary, but that adds too much complexity to Vault due to the conditional code required for this.

schultz-is (approved):

PR #178, opened 2023-07-18
Use NewCredentialsConfig

bdwyertech:
This should resolve issues with IRSA not working


jimlambrt:
Given https://github.com/hashicorp/vault/pull/21951 
Is this PR still relevant PR?

PR #179, opened 2023-07-19
Add a codeowners file

sgmiller:
For now, it assigns the entire codebase to both Vault Crypto and Boundary.


jefferai (approved):
Seems fine. I might create a boundary-crypto team at some point or something but for now it works.

Actually, may be worth adding me directly? Unsure. If you think so add it, approved either way.

PR #183, opened 2023-09-01
Work around an azure HTTP/2 bug

sgmiller:
Azure's client by default doesn't health check HTTP/2 keepalive conns, and instead lets the os close them after inactivity if they become unhealthy or are closed unexpectedly.  This can take 15 minutes.  If too many idle conns get hung in this state the client becomes unusable.  This workaround was provided by Microsoft as they investigate the issue.

Issue: https://github.com/Azure/azure-sdk-for-go/issues/21346
Workaround is frrom that issue but backported to the older Azure client.


jimlambrt (dismissed):
It appears you forgot to add your unit test commits to the PR.  Would you mind adding them?

sgmiller:
Unfortunately this can't be unit tested.  It relies on HTTP/2 connections being dropped on the Azure side, which is very intermittent.  The rest of the change is covered by the acceptance tests that existed already.

victorr (approved):
👍

victorr (approved):

jimlambrt:
I'm a bit skeptical that the existing vault unit tests asserted anything about the custom transport you've added.  

I've pushed some unit tests which do some minimal assertions about the fix.

jefferai (approved):
Obviously this mostly relies on MSFT being correct in their given workaround but seems fine... any chance you have figured out a repeated way to trigger it to see if this actually helps?

sgmiller:
Unfortunately no.  I might be able to do it with iptables but not great for unit testing.

sgmiller:
> I'm a bit skeptical that the existing vault unit tests asserted anything about the custom transport you've added.
> 
> I've pushed some unit tests which do some minimal assertions about the fix.

Okay, that's a useful thing to verify.  Thanks.

PR #186, opened 2023-09-14
Key id needs to be more specific than just the version in order to use Seal HA with multiple transit seals

sgmiller:
Don't merge yet, need to think about backwards compatibility


sgmiller:
Closing in favor of a different approach

PR #187, opened 2023-09-15
Provide a keyId prefix as an option to transit wrappers

sgmiller:
This allows Vault's seal subsystem to distinguish between two transit wrappers by passing a distinct prefix for each, but doesn't affect keyId for other consumers.


jefferai (approved):

jimlambrt (approved):
Ty

PR #207, opened 2023-11-10
Allow plugin package to be compiled under WASM

swenson:
Note that it certainly doesn't *work* under WASM, but at least it should not cause compile errors.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -10,7 +10,7 @@ jobs:
   build:
     strategy:
       matrix:
-        go: ["1.21", "1.20", "1.19"]
+        go: ["1.21", "1.20"]
         platform: [ubuntu-latest] # can not run in windows OS
     runs-on: ${{ matrix.platform }}
 
@@ -43,7 +43,7 @@ jobs:
   sqlite:
     strategy:
       matrix:
-        go: ["1.21", "1.20", "1.19"]
+        go: ["1.21", "1.20"]
         platform: [ubuntu-latest] # can not run in windows OS
     runs-on: ${{ matrix.platform }}
 
@@ -77,7 +77,7 @@ jobs:
             "postgres:11",
             "postgres:10",
           ]
-        go: ["1.21", "1.20", "1.19"]
+        go: ["1.21", "1.20"]
         platform: [ubuntu-latest] # can not run in macOS and Windows
     runs-on: ${{ matrix.platform }}
 
@@ -119,7 +119,7 @@ jobs:
   plugin-transit:
     strategy:
       matrix:
-        go: ["1.21", "1.20", "1.19"]
+        go: ["1.21", "1.20"]
         platform: [ubuntu-latest] # can not run in macOS and Windows
     runs-on: ${{ matrix.platform }}
 
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -30,15 +30,15 @@ jobs:
           key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('tests/go.mod') }}
 
       - name: Build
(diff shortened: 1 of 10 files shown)


.github/workflows/go.yml
```diff
@@ -32,6 +32,9 @@ jobs:
       - name: Build
         run: CGO_ENABLED=0 find . -name go.mod -execdir go build ./... \;
 
+      - name: WASM Build
```
johanbrandhorst, line 35: This is ambiguous with the new `wasip1` target, could we make it more explicit?
```suggestion
      - name: JS/WASM Build
```

swenson:
(the builds are failing on 1.17 but are passing anyway because I think the `find` is hiding the error code?)

jefferai:
Think you could fix the test?

swenson:
Sure, let me take a look at `find` and see if I can fix it.

swenson:
@jefferai build and tests fixed and will actually fail now when appropriate :)

jefferai:
Awesome thanks! I will defer to @johanbrandhorst on hitting Approve on this since I've not dealt with wasm and he knows a lot about it :-)

.github/workflows/go.yml
```diff
@@ -30,17 +30,20 @@ jobs:
           key: ${{ runner.os }}-go-${{ matrix.go }}-${{ hashFiles('tests/go.mod') }}
 
       - name: Build
-        run: CGO_ENABLED=0 find . -name go.mod -execdir go build ./... \;
+        run: find . -name go.mod -execdir sh -c 'pwd && CGO_ENABLED=0 go build ./... || exit 1' sh {} +
```
johanbrandhorst, line 33: Can't this just be
```suggestion
        run: CGO_ENABLED=0 go build ./...
```
? Does it need to build independently in each module?

.github/workflows/go.yml
```diff
@@ -30,17 +30,20 @@ jobs:
...
+
+      - name: JS/WASM Build
+        run: find . -name go.mod -not -path "*examples*" -execdir sh -c 'pwd && GOOS=js GOARCH=wasm CGO_CENABLED=0 go build ./... || exit 1' sh {} +
```
johanbrandhorst, line 36: This seems wrong. I don't think you should need to set CGO_ENABLED=0 for wasm anyway.
```suggestion
        run: find . -name go.mod -not -path "*examples*" -execdir sh -c 'pwd && GOOS=js GOARCH=wasm CGO_ENABLED=0 go build ./... || exit 1' sh {} +
```

.github/workflows/go.yml
swenson: It does. `go build ./...` does not seem to build the the sub-directories.

You can tell because `GOOS=js GOARCH=wasm go build ./...` works, but `cd plugin && GOOS=js GOARCH=wasm go build ./...` fails (correctly).

.github/workflows/go.yml
swenson: Ha, yep, that was my bad. And it's not necessary I don't think.

johanbrandhorst (approved):

swenson:
Thanks!

jefferai:
A couple of things...

One -- I can't see any reason for the updates the various submodules that only contain go.mod/go.sum updates. Is there a reason, and if not, can they be reverted?

Two -- not all tests are passing, is that expected?

swenson:
@jefferai 

1. `hashicorp/go-sockaddr` >=1.0.6 is needed to compile to WASM (that is the release after I did https://github.com/hashicorp/go-sockaddr/pull/55)

2. The only test that isn't passing is `buf_if_merged`, and I honestly haven't investigated what that test is. It appears to be misconfigured (missing token?).

swenson:
Thanks!

PR #224, opened 2024-01-26
[azurekeyvault] remove autorest dependencies

kschoche:
* Removes deprecated autorest library calls from azurekeyvault
* Removes dead functions
* Fixes a bug whereby we do not pass in the cloudConfigOptions to `azidentity.NewDefaultAzureCredential()` and `azidentity.NewClientSecretCredential()`


wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -180,7 +193,11 @@ func (v *Wrapper) SetConfig(ctx context.Context, opt ...wrapping.Option) (*wrapp
 			if keyInfo.Key == nil {
 				return nil, errors.New("no key information returned")
 			}
-			v.currentKeyId.Store(ParseKeyVersion(to.String((*string)(keyInfo.Key.KID))))
+			keyID := ""
```
jimlambrt, line 196: didn't the previous code accomplish the same thing given that to.String(...)
returns an empty string if it's given a ni?

jimlambrt (changes requested):
The options you're removing all appear to be valid options.  Can you explain a bit why you're removing them?

wrappers/azurekeyvault/azurekeyvault.go
kschoche: Yup, the goal is to remove autorest based libraries as they are deprecated.

kschoche:
> The options you're removing all appear to be valid options. Can you explain a bit why you're removing them?
Thanks!
Maybe I'm misunderstanding, are you referring to the unused functions that were removed in `options.go` ?
I can definitely revert that bit if I've misunderstand where they are used.

jimlambrt:
See all the opts.* in `SetConfig(...)` to see how they're used.

kschoche:
> See all the opts.* in `SetConfig(...)` to see how they're used.

Thanks for pointing that out, my IDE had it all flagged as unused, my mistake for not checking further!

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -365,3 +387,31 @@ func ParseKeyVersion(kid string) string {
 	keyVersionParts := strings.Split(kid, "/")
 	return keyVersionParts[len(keyVersionParts)-1]
 }
+
+func cloudConfigFromName(name string) (cloud.Configuration, error) {
+	configs := map[string]cloud.Configuration{
```
jefferai, line 392: Can this be a simple switch statement on strings.ToUpper(name)?

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -365,3 +387,31 @@ func ParseKeyVersion(kid string) string {
...
+	name = strings.ToUpper(name)
+	c, ok := configs[name]
+	if !ok {
+		return c, fmt.Errorf("err: no cloud configuration matching the name %q", name)
+	}
+	return c, nil
+}
+
+func keyVaultDNSSuffixFromName(envName string) (string, error) {
+	urls := map[string]string{
```
jefferai, line 406: Same here?

Alternately it could be one function where the env names map to structs containing both the cloud config and dns suffix and the calling function can just choose which it wants from the returned struct.

kschoche: Yeah definitely can do that, thanks for the suggestion!

kschoche: Fixed in 4d94f4c9288da3d640e9bcd16137c337382b650e

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -131,12 +141,15 @@ func (v *Wrapper) SetConfig(ctx context.Context, opt ...wrapping.Option) (*wrapp
...
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
-	v.environment.KeyVaultDNSSuffix = azResource
+	v.keyVaultDNSSuffix = azResource
 	v.resource = "https://" + azResource + "/"
-	v.environment.KeyVaultEndpoint = v.resource
+	v.keyVaultEndpoint = v.resource
```
austingebauer, line 152: `keyVaultEndpoint` looks unused after it's set here? I wonder if we're supposed to set the endpoint on the `cloudConfig` similar to the environment before.

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -365,3 +387,16 @@ func ParseKeyVersion(kid string) string {
 	keyVersionParts := strings.Split(kid, "/")
 	return keyVersionParts[len(keyVersionParts)-1]
 }
+
+func cloudConfigAndKeyVaultDNSSuffixFromName(name string) (cloud.Configuration, string, error) {
+	switch name {
```
austingebauer, line 392: Should this be `switch strings.ToUpper(name) {`? The environment values are typically provided in pascal case like AzurePublicCloud, AzureChinaCloud, ...

wrappers/azurekeyvault/azurekeyvault_acc_test.go
```diff
@@ -71,7 +71,7 @@ func TestAzureKeyVault_IgnoreEnv(t *testing.T) {
 	require.Equal(t, config["tenant_id"], s.tenantID)
 	require.Equal(t, config["client_id"], s.clientID)
 	require.Equal(t, config["client_secret"], s.clientSecret)
-	require.Equal(t, config["environment"], s.environment.Name)
+	require.Equal(t, config["environment"], azure.PublicCloud.Name)
```
austingebauer, line 74: nit: If you provide "AzurePublicCloud" in place of `azure.PublicCloud.Name` in the test code we can drop the Azure/go-autorest dependency completely.

PR #250, opened 2024-04-08
Fix the fact that some GCP env vars are immune to WithDisallowEnvVars

sgmiller:


victorr (approved):
👍

wrappers/gcpckms/gcpckms_acc_test.go
```diff
@@ -23,7 +23,6 @@ const (
 // TestGcpKeyIdAfterConfig will test the result of calling the wrapper's KeyId()
 // after it's configured with various options
 func TestGcpKeyIdAfterConfig(t *testing.T) {
-	t.Parallel()
```
stevendpclark, line 26: This was removed as it didn't play nice with the new test case disabling env-vars

wrappers/gcpckms/gcpckms_acc_test.go
```diff
@@ -91,10 +90,55 @@ func TestGcpKeyIdAfterConfig(t *testing.T) {
...
+		t.Fatalf("err: %s", err.Error())
+	}
+
+	if !reflect.DeepEqual(input, pt) {
+		t.Fatalf("expected %s, got %s", input, pt)
+	}
 }
 
 func TestGcpCkmsSeal(t *testing.T) {
+	t.Setenv(EnvGcpCkmsWrapperProject, "") // Make sure at least one required value is not set.
```
stevendpclark, line 140: This test worked correctly if it ran by itself but running them all together, if another test ran first and set the appropriate vars would fail as we wouldn't get the failure we expected.

jefferai (approved):

PR #259, opened 2024-05-10
wrappers/transit: support context cancelation

johanbrandhorst:
This makes the transit client respect context cancelation, which is a critical feature of any I/O API.

Build failure is fixed in https://github.com/hashicorp/go-kms-wrapping/pull/260.


jimlambrt (approved):
Ty!

PR #262, opened 2024-06-12
add naver cloud kms wrapper

jayground8:
Hello, It is to support Naver Cloud KMS.

[This is a documentation for Naver Cloud KMS](https://guide.ncloud-docs.com/docs/en/kms-overview)

I tested `ncloudkms_test.go` with valid access key and secret key.


jayground8:
Hello @jefferai 

I would like to support Naver Cloud KMS for Vault auto unseal. Can I get any update for this PR?

Thank you.

PR #272, opened 2025-01-10
Support parseutil.ParsePath for sensitive values in wrapper configs

sgmiller:
Quite a few operators have asked for the ability to pull sensitive values into seal/managed key configs from separate files, feeling env vars or direct config isn't secure.  ParsePath gives them the option to get values from a separate file while being non-invasive to existing configs.


sgmiller:
Will subsequently update the go.mod deps to a tagged version of go-kms-wrapping/v2

options.go
```diff
@@ -154,3 +155,14 @@ func WithoutHMAC() Option {
 		})
 	}
 }
+
+func ParsePaths(fields ...*string) error {
+	for i := 0; i < len(fields); i++ {
+		if newVal, err := parseutil.ParsePath(*fields[i]); err != nil {
```
stevendpclark, line 161: Shouldn't this trap `errors.Is(err, ErrNotAUrl)` and not return it leaving whatever the field value was previously?

I'm just wondering if we considered with these little tidbits but all three are me trying to poke holes into and I don't feel strongly about them.

1.  If we were given any values that started or ended with spaces they will no longer work as ParsePath will trim them.
2. It's a little unfortunate that with `env://` values we don't get an error back saying it didn't exist which may be hard for operators to diagnose connection issues in the future, but we do for file based schemas
3. Obviously if the values being passed in had values beginning with file:// or env:// that were for some reason meant to be passed to the underlying wrapper there is no way out now.

sgmiller:
Updated this to just be the main package level changes.  Next step if approved/merged is to tag gkw, then update the wrappers to use it against a tagged dep.

stevendpclark (dismissed):
👍

PR #275, opened 2025-01-22
Use ParsePaths to source sensitive wrapper config values from Env/File if desired

sgmiller:
Hopefully I identified which config elements are "sensitive" per wrapper correctly, that's definitely something to get a second pair of eyes on.


stevendpclark (commented):
Looks good 👍 Two small comments before approval.

stevendpclark (approved):

jefferai (approved):

PR #277, opened 2025-01-30
Facilitate use of User Managed Identities

sgmiller:
User managed identities in Azure are referenced only by client_id (if at
all), and don't use a client_secret. When client_secret is absent in wrapper
config, attempt to use this form of credential.

Fixes https://hashicorp.atlassian.net/browse/VAULT-32283 and https://hashicorp.atlassian.net/browse/VAULT-33015


stevendpclark (approved):
LGTM (I can't think of any way to unit test this that would provide any real value), and I looked through the changes for both Azure dependency updates and nothing stood out. 

The go.sum could use a cleanup through `go mod tidy`

sgmiller:
> LGTM (I can't think of any way to unit test this that would provide any real value), and I looked through the changes for both Azure dependency updates and nothing stood out.
> 
> The go.sum could use a cleanup through `go mod tidy`

Yeah, testing it in Azure was a pain too, even w/ terraform, but I did set up one user managed identity, tested it with and without client_id (and no client_secret), then two, one with permissions to access key vault and one without, so I could verify that the client_id was taking effect to select each one.

jefferai (approved):

PR #278, opened 2025-02-04
Update golang.org/x/net to 0.34.0 across all modules

stevendpclark:
- Resolve a bunch of security warnings around "Non-linear parsing of case-insensitive content" from the use of the golang.org/x/net package


victorr (approved):
👍

jefferai (approved):

PR #280, opened 2025-02-28
Fix incompatibility between workload identity and managed identity auth

sgmiller:
The default credential provider _does_ handle managed service identities, but can only get the client id from an env var.  Supply it via env var (as there is no way in the Azure API to do so), but rely on the default credential provider to decide which auth pattern to use.

### DIFF
--- a/wrappers/azurekeyvault/azurekeyvault.go
+++ b/wrappers/azurekeyvault/azurekeyvault.go
@@ -300,7 +300,8 @@ func (v *Wrapper) getKeyVaultClient(withCertPool *x509.CertPool) (*azkeys.Client
 			return nil, fmt.Errorf("failed to get client secret credentials %w", err)
 		}
 	case v.clientID != "":
-		// Set an env var with the client id, and let the default credential provider work through the possibilities
+		// This could be a managed identity auth, so supply the default credential provider with clientId and let it
+		// figure it out.
 		os.Setenv(EnvAzureClientId, v.clientID)
 		fallthrough
 	// By default let Azure select existing credentials
--- a/wrappers/azurekeyvault/azurekeyvault.go
+++ b/wrappers/azurekeyvault/azurekeyvault.go
@@ -34,6 +34,7 @@ const (
 
 	EnvAzureKeyVaultWrapperKeyName = "AZUREKEYVAULT_WRAPPER_KEY_NAME"
 	EnvVaultAzureKeyVaultKeyName   = "VAULT_AZUREKEYVAULT_KEY_NAME"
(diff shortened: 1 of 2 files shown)


stevendpclark (approved):

vinay-gopalan (approved):

sgmiller:
Fixes https://github.com/hashicorp/vault/issues/29717

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -299,14 +303,28 @@ func (v *Wrapper) getKeyVaultClient(withCertPool *x509.CertPool) (*azkeys.Client
...
-		// Try a managed service credential with a specified client id
-		clientID := azidentity.ClientID(v.clientID)
-		cred, err = azidentity.NewManagedIdentityCredential(&azidentity.ManagedIdentityCredentialOptions{
-			ID: clientID,
-		})
+		// Some hoops to jump through to make sure two wrappers being setup at the same time don't step on the
+		// env var
+		managedClientIdLock.Lock()
+		oldVal, found := os.LookupEnv(EnvAzureClientId)
+		unlock := func() {
```
jimlambrt, line 310: any reason to not use a `defer`?

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -287,6 +289,8 @@ func (v *Wrapper) buildBaseURL() string {
 	return fmt.Sprintf("https://%s.%s/", v.vaultName, v.environment.KeyVaultDNSSuffix)
 }
 
+var managedClientIdLock sync.Mutex
```
jimlambrt, line 292: Can this be moved into the Wrapper struct?

jimlambrt (changes requested):
Sorry, I have a couple questions.

wrappers/azurekeyvault/azurekeyvault.go
sgmiller: Yeah, the defer takes the execution down into the part in the code below where a connection is made, which takes several seconds sometimes.  We want to hold the lock only as long as it takes to initialize the credential provider.

wrappers/azurekeyvault/azurekeyvault.go
sgmiller: No, it has to be global, as the race is between more than one instance of Wrapper being initialized in Vault, when used with Seal High Availability.

jimlambrt: Ty!

jimlambrt (approved):

PR #301, opened 2026-03-06
Add a "WithoutEnvelope" option to wrappers for raw Encrypt/Decrypt support

sgmiller:
By default many of the wrappers use Envelope encryption during
encrypt/decrypt.  This is helpful for the common case like seal wrapping
where we want to support arbitrary sized plaintexts without worrying about
the underlying cipher's limits.  However, these days, wrappers are used as a
general interface for external cryptography.  Some features, like PKI
signing, require more low level operations.  An example is SCEP, which uses
the CA private key for encryption and decryption as well.  Another
is Transit managed keys, which support cryptographic primitive operations
that may need to interact with other systems.  In these cases we need the
simple encrypt/decrypt without an envelope.

This PR adds a new global wrapper option, WithoutEnvelope(), which if
present requests that the wrapper use direct encryption/decryption.

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.


stevendpclark (dismissed):

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -255,34 +295,69 @@ func (v *Wrapper) Decrypt(ctx context.Context, in *wrapping.BlobInfo, opt ...wra
 		return nil, errors.New("given input for decryption is nil")
 	}
 
+	// Default to mechanism used before key info was stored
 	if in.KeyInfo == nil {
-		return nil, errors.New("key info is nil")
+		in.KeyInfo = &wrapping.KeyInfo{
```
stevendpclark: I think we'd want to keep the existing behavior if `in.KeyInfo == nil` to return an error as there are other fields we need that aren't set such as `KeyId`, `WrappedKey`.

And the default for Mechanism will just work with the iota

wrappers/azurekeyvault/azurekeyvault.go
```diff
@@ -255,34 +295,69 @@ func (v *Wrapper) Decrypt(ctx context.Context, in *wrapping.BlobInfo, opt ...wra
...
-	algo := azkeys.JSONWebKeyEncryptionAlgorithmRSAOAEP256
-	params := azkeys.KeyOperationsParameters{
-		Algorithm: &algo,
-		Value:     wrappedBytes,
-	}
+	var plaintext []byte
+	switch in.KeyInfo.Mechanism {
+	case AzureKeyVaultEncrypt:
+		// Unwrap the plaintext directly
+		wrappedBytes, err := base64.RawURLEncoding.DecodeString(string(in.Ciphertext))
```
stevendpclark: Do we need this for the new method? Feels like a carry over from the previous EnvelopeEncryption

wrappers/gcpckms/gcpckms.go
```diff
@@ -220,35 +220,64 @@ func (s *Wrapper) Encrypt(ctx context.Context, plaintext []byte, opt ...wrapping
...
-		KeyInfo: &wrapping.KeyInfo{
-			Mechanism: GcpCkmsEnvelopeAesGcmEncrypt,
-			// Even though we do not use the key id during decryption, store it
-			// to know exactly what version was used in encryption in case we
-			// want to rewrap older entries
-			KeyId:      resp.Name,
-			WrappedKey: resp.Ciphertext,
-		},
-	}
+	var ret *wrapping.BlobInfo
```
stevendpclark, line 228: Missing the corresponding changes to Decrypt for GCP

wrappers/huaweicloudkms/huaweicloudkms.go
```diff
@@ -141,29 +176,53 @@ func (k *Wrapper) Decrypt(_ context.Context, in *wrapping.BlobInfo, opt ...wrapp
 		return nil, fmt.Errorf("given input for decryption is nil")
 	}
 
-	// KeyId is not passed to this call because HuaweiCloud handles this
-	// internally based on the metadata stored with the encrypted data
-	plainText, err := k.client.decrypt(string(in.KeyInfo.WrappedKey))
-	if err != nil {
-		return nil, fmt.Errorf("error decrypting data encryption key: %w", err)
+	// Default to mechanism used before key info was stored
```
stevendpclark: Don't think this is needed, or we could keep the `in.KeyInfo == nil` and return an error if it is

wrappers/ibmkp/ibmkp.go
```diff
@@ -194,29 +225,49 @@ func (k *Wrapper) Decrypt(ctx context.Context, in *wrapping.BlobInfo, opt ...wra
 		return nil, errors.New("given input for decryption is nil")
 	}
 
+	// Default to mechanism used before key info was stored
 	if in.KeyInfo == nil {
-		return nil, errors.New("key info is nil")
+		in.KeyInfo = &wrapping.KeyInfo{
```
stevendpclark: Same should keep the error being returned here

wrappers/ocikms/ocikms.go
```diff
@@ -205,33 +257,67 @@ func (k *Wrapper) Decrypt(ctx context.Context, in *wrapping.BlobInfo, opt ...wra
...
-	if err != nil {
-		return nil, fmt.Errorf("error base64 decrypting data: %w", err)
-	}
-	envInfo := &wrapping.EnvelopeInfo{
-		Key:        envelopeKey,
-		Iv:         in.Iv,
-		Ciphertext: in.Ciphertext,
+	// Default to mechanism used before key info was stored
+	if in.KeyInfo == nil {
+		in.KeyInfo = &wrapping.KeyInfo{
```
stevendpclark: Similar an error should be returned here

wrappers/tencentcloudkms/tencentcloudkms.go
```diff
@@ -192,28 +228,56 @@ func (k *Wrapper) Decrypt(_ context.Context, in *wrapping.BlobInfo, opt ...wrapp
...
 
-	input := kms.NewDecryptRequest()
-	input.CiphertextBlob = common.StringPtr(string(in.KeyInfo.WrappedKey))
-
-	output, err := k.client.Decrypt(input)
-	if err != nil {
-		return nil, fmt.Errorf("error decrypting data encryption key: %w", err)
+	// Default to mechanism used before key info was stored
+	if in.KeyInfo == nil {
+		in.KeyInfo = &wrapping.KeyInfo{
```
stevendpclark: Should return error

wrappers/huaweicloudkms/huaweicloudkms.go
sgmiller: Looks like adding the keyInfo check is good anyway. Above would be a nil pointer.

wrappers/awskms/go.mod
```diff
@@ -2,16 +2,17 @@ module github.com/hashicorp/go-kms-wrapping/wrappers/awskms/v4
 
 go 1.24.0
 
+replace github.com/hashicorp/go-kms-wrapping/v2 => ../..
```
stevendpclark: This should be cleaned up.

wrappers/gcpckms/gcpckms.go
```diff
@@ -220,35 +220,64 @@ func (s *Wrapper) Encrypt(ctx context.Context, plaintext []byte, opt ...wrapping
...
+			KeyInfo: &wrapping.KeyInfo{
+				Mechanism: GcpCkmsEnvelopeAesGcmEncrypt,
+				// Even though we do not use the key id during decryption, store it
+				// to know exactly what version was used in encryption in case we
+				// want to rewrap older entries
+				KeyId:      resp.Name,
+				WrappedKey: resp.Ciphertext,
+			},
+		}
+	}
```
stevendpclark, line 280: Still missing the Decrypt here for GCP.

stevendpclark: Interesting the decrypt implementation was already present, nvm, I had just assumed since it wasn't appearing in the PR it hadn't been implemented.

stevendpclark (approved):

PR #335, opened 2026-08-18
Add RSA encryption padding option

DeLuci:
## PCI review checklist

Adds an encryption option to allow RSA oeap (sha1/sha2256) or RSA Pkcs1v15 encryption.

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.


github.com.hashicorp.go.kms.wrapping.v2.types.proto
```diff
@@ -149,6 +149,9 @@ message Options {
 
   // WithoutEnvelope specifies that encryption should be over the plaintext rather than using an envelope encryption pattern
   bool without_envelope = 110;
+
+  // WithEncryptionAlgorithm specifies the RSA encryption algorithm to use (optional)
+  EncryptionAlgorithm with_encryption_algorithm = 120;
```
sgmiller: Since this is specific to RSA, lets make that explicit in the field name, maybe RSAEncryptionPadding

options.go
```diff
@@ -166,6 +166,16 @@ func WithoutEnvelope(rawEncryption bool) Option {
 	}
 }
 
+// WithEncryptionAlgorithm specifies the RSA encryption algorithm (padding scheme + hash) to use
+func WithEncryptionAlgorithm(alg EncryptionAlgorithm) Option {
```
sgmiller: Ditto the helper functions.  This sounds like it would apply to all encryption but it's really RSA only.

sgmiller (changes requested):

sgmiller (approved):

victorr (approved):
👍

PR #348, opened 2026-09-02
TencentCloud KMS wrapper: CAM role support, endpoint override, and env var fix

flyfishli:
## Description

Adds two capabilities to the TencentCloud KMS wrapper and fixes one
inconsistency with the other cloud wrappers.

### 1. CAM role support (AssumeRole)

The wrapper could only use static credentials. It can now assume a CAM role
via STS and use the resulting temporary credentials for KMS, so the configured
`access_key`/`secret_key` only need `sts:AssumeRole` rather than direct KMS
permissions. `role_arn` may reference a role in another account, which enables
cross-account KMS access.

New settings (config map / env var / option func):

| Config key | Env var |
|---|---|
| `role_arn` | `TENCENTCLOUD_ROLE_ARN` |
| `role_session_name` | `TENCENTCLOUD_ROLE_SESSION_NAME` |
| `role_external_id` | `TENCENTCLOUD_ROLE_EXTERNAL_ID` |
| `role_duration_seconds` | `TENCENTCLOUD_ROLE_DURATION_SECONDS` |

`role_duration_seconds` is validated client-side against the range STS accepts
(900-43200s). When unset, the service default of 7200s applies.

### 2. Endpoint override

The wrapper always used the default public KMS domain, so it could not be used
where KMS must be reached through a VPC endpoint or in a dedicated cloud
environment. `endpoint` / `TENCENTCLOUD_KMS_ENDPOINT` / `WithEndpoint()` now
allow overriding it. This mirrors `endpoint` in awskms and `domain` in
alicloudkms.

Note that the KMS and STS clients are deliberately given separate client
profiles: the endpoint override must apply to KMS only, since sharing a single
profile would send AssumeRole calls to the KMS endpoint.

### 3. Fix: honor `WithDisallowEnvVars`

Unlike alicloudkms, awskms and the others, this wrapper read `os.Getenv`
unconditionally and ignored `WithDisallowEnvVars`. Since Vault sets that option
and resolves env vars into the config map itself, environment values were
consulted a second time and silently took precedence over explicit
configuration. Every lookup is now guarded, matching the other wrappers.

## Backwards compatibility

- All new settings are optional; when `role_arn` is unset no STS call is made
  and the credential handling is byte-for-byte the same as before.
- Changes are confined to `wrappers/tencentcloudkms/`; no shared code is touched.
- The `WithDisallowEnvVars` fix is a behaviour change for callers that both set
  that option *and* rely on env vars leaking through. That combination is
  contradictory, and the other wrappers already behave this way.

## Testing

The pre-existing tests are all gated behind `VAULT_ACC` and therefore skipped
by default, so this adds `tencentcloudkms_role_test.go` with tests that need
neither credentials nor network access, covering:

- `role_duration_seconds` validation on both the config map and option paths
- role and endpoint settings from config map, option funcs, and env vars
- `WithDisallowEnvVars` being honored for credential, role, and endpoint settings
- AssumeRole request assembly (session name default, omission of unset fields)
- KMS/STS client profile isolation
- a regression guard asserting nothing changes when no role is configured

Package coverage goes from effectively 0% to ~53%.

## PCI review checklist

- [x] I have documented a clear reason for, and description of, the change I am making.

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

  Reverting the pull request is sufficient; there are no migrations or persisted
  state changes involved.

- [x] If applicable, I've documented the impact of any changes to security controls.

  This change adds an authentication path (STS AssumeRole) that lets operators
  grant KMS access to a role instead of embedding long-lived credentials with
  direct KMS permissions, which reduces the blast radius of a leaked key.
  `role_external_id` is supported for confused-deputy protection on
  cross-account access. The `WithDisallowEnvVars` fix removes a case where
  environment variables could silently override explicit configuration.

### DIFF
--- a/wrappers/tencentcloudkms/go.mod
+++ b/wrappers/tencentcloudkms/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/hashicorp/go-kms-wrapping/v2 v2.0.24
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common v1.0.604
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/kms v1.0.604
+	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/sts v1.0.604
 )
 
 require (
--- a/wrappers/tencentcloudkms/go.sum
+++ b/wrappers/tencentcloudkms/go.sum
@@ -24,6 +24,8 @@ github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common v1.0.604 h1:BiuY
 github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common v1.0.604/go.mod h1:7sCQWVkxcsR38nffDW057DRGk8mUjK1Ing/EFOK8s8Y=
 github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/kms v1.0.604 h1:LItsuG1cD+8e/9Z5MBNci2WJpYeNI0j00FN8UU5A3xw=
 github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/kms v1.0.604/go.mod h1:0NPGaAgj1qcvy3w0wEVhe3hMmDB0ptM+CqzXDu9r658=
+github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/sts v1.0.604 h1:4AulxOLAOI/SuBxBwPPiwPE9OeGM2e1UMZNsbfhizqQ=
+github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/sts v1.0.604/go.mod h1:w+DBpcgJn2iFZURqvsmkDHbd1w5/q0YwSPDSt2+9MRY=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543 h1:E7g+9GITq07hpfrRu66IVDexMakfv52eLZ2CXBWiKr4=
 golang.org/x/xerrors v0.0.0-20191204190536-9bdfabe68543/go.mod h1:I/5z698sn9Ka8TeJc9MKroUUfqBBauWjQqLJ2OPfmY0=
 google.golang.org/protobuf v1.33.0 h1:uNO2rsAINq/JlFpSdYEKIZ0uKD/R9cpdv0T+yoGwGmI=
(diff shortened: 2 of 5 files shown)


flyfishli:
Hi @sgmiller @ncabatoff — could either of you take a look when there's a chance?
Happy to adjust anything.

### Context / motivation

We're preparing a PR against `hashicorp/vault` to wire this wrapper into the seal
configuration, so Tencent Cloud KMS can be used for auto-unseal the same way AWS,
Azure, GCP and Alibaba Cloud already can. While doing that integration we found
the wrapper was missing a few things needed to make it usable in practice, which
is what this PR addresses.

Notably, **`9ad1c85` is a bug that affects Vault directly**: unlike alicloudkms
and awskms, this wrapper called `os.Getenv` unconditionally and ignored
`WithDisallowEnvVars`. Vault sets that option and resolves environment variables
into the config map itself, so environment values were read a second time and
silently took precedence over an explicitly supplied seal configuration. This
commit simply brings the wrapper in line with the others.

The remaining commits close the feature gap with awskms/alicloudkms so the seal
stanza can express what operators actually need — assuming a CAM role instead of
embedding long-lived credentials, and reaching KMS through a VPC endpoint or in a
dedicated cloud environment.

### The five commits are self-contained

| Commit | What it does |
|---|---|
| `476cab8` | Adds CAM role assumption (STS AssumeRole) |
| `9ad1c85` | **Bugfix**: honor `WithDisallowEnvVars` |
| `1780093` | Unit tests (no credentials or network needed) |
| `088343d` | Adds an endpoint override |
| `0d81ab0` | Uses the public KMS domain in examples |

### Notes for review

- **Scope**: everything is inside `wrappers/tencentcloudkms/`. No shared code, no
  other wrapper, and no public API of the core module is touched.
- **Backwards compatibility**: all new settings are optional. With `role_arn`
  unset, no STS call is made and credential handling is identical to before.
- **Endpoint/STS isolation**: the KMS and STS clients deliberately get separate
  client profiles, so an endpoint override never redirects AssumeRole calls.
  Covered by `TestNewClientProfile`.
- **Tests**: the pre-existing tests are all gated behind `VAULT_ACC` and skipped
  by default, so this package had effectively no coverage. The new tests require
  neither credentials nor network access and bring it to ~53%.

Verified end to end against a real Tencent Cloud KMS key with a locally built
Vault: initialisation, auto-unseal after restart, and both same-account and
cross-account role assumption all work. Seal migration (Shamir → KMS, and KMS key
rotation) was tested as well.

Once this lands and a new module version is tagged, we'll open the corresponding
PR on `hashicorp/vault`.

This contribution is submitted on behalf of Tencent Cloud KMS
(https://cloud.tencent.com/product/kms).

